PR #104, opened 2014-09-23
changed priority colors to more contrast

PanderMusubi:


dmfs:
With the upcoming release we've changed the colors to three colors from the material color palette ( #4caf50, #f44336 and #ffc107), see http://www.google.com/design/spec/style/color.html. The contrast is probably not as high as with the "pure" colors you've picked, but I think they play nicely with the overall color scheme and contrast is still very good. Hope you like it.

PR #137, opened 2015-05-01
Update French translation

Ezelty:
Hello,
I improved the French translation. I made some changes and added missing strings.
Greetings


dmfs:
Thanks! That's great. I'll review and merge during the next few days.

PR #144, opened 2015-06-13
App Install Location

SweetLion:
The internal memory of my device is quite small, this PR should allow users to easily move the app to the microSD.

"To allow the system to install your application on the external storage, modify your manifest file to include the android:installLocation attribute in the element, with a value of either "preferExternal" or "auto"." (source: https://developer.android.com/guide/topics/data/install-location.html)

Thank you very much for your awesome app :)

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.dmfs.tasks"
+    android:installLocation="auto"
     android:versionCode="84"
     android:versionName="1.1.6" >
 


dmfs:
I understand the problem, but I'm not sure if this solution is really a good idea. The problem is that other apps (the task sync apps) depend on the availability of this app (of the TaskProvider in particular). Since there is no guarantee that the SD card is always mounted (and the TaskProvider always present) unexpected things might happen.
On many Android 2.x devices the SD card gets unmounted while the phone is connected to a computer (so the computer can mount it).

SweetLion:
I see your point. It is correct that the phone cannot access the microSD while connected to a PC (e. g. to transfer pictures). I suppose it affects every Android version. I am using 2.3.6 but I know from a Android 4 (I think KitKat) user that it is the same there. When the phone connects to the PC the apps on the microSD shut down and disappear from the app drawer. After disconnecting it the apps reappear.
Making sure that the TaskProvider etc. restarted after disconnecting the phone from the PC is a bit tricky. 

Perhaps it is safest to close this PR. 

Thank you for your interest and thinking about it.

PR #166, opened 2015-10-03
Add Ukrainian lang

burunduk:


dmfs:
Thanks! We're currently verifying the translation.
Is there any specific reason why you've put the plurals into a separate file?

burunduk:
No. I copy and translate this file from apk.

dmfs:
I've just merged it locally. The translation will be present in the upcoming release. I'll probably move the plurals back into the strings to ensure all strings files have the same structure and order of elements, which makes them easier to compare.

PR #174, opened 2015-10-31
Update Serbian

pejakm:


dmfs:
Thanks a lot. I've just merged it locally. The updates will be present in the next release. I've just reverted the app name to OpenTasks. We don't localize the name of the app itself.

pejakm:
Thanks, but could you please consider leaving localized name at least for Serbian? Since we are using Cyrillic script, translation is considered unfinished if name isn't localized. General consensus among Serbian translation community is that we either translate or transcribe app name.

dmfs:
Really? How about "Android", "Google" or "Facebook"? Do you use transcribed versions of these names?
"OpenTasks" is not the title underneath the launcher icon (for that we use the localized version of "tasks"), so you won't see "OpenTasks" in your app drawer.

pejakm:
> How about "Android", "Google" or "Facebook"? Do you use transcribed versions of these names?

We can't influence on companies and proprietary projects, but in common communication (including TV, newspapers, etc.) if the text is in Cyrillic script those names are usually transcribed („Андроид“, „Гугл“, „Фејсбук“). „Usually“ means „if authors strictly follow language rules“. (As for „Android“: in AOSP it left in Latin script, I'm not sure why, but AFAIK every other FLOSS project uses transcribed name.)

> "OpenTasks" is not the title underneath the launcher icon (for that we use the localized version of "tasks"), so you won't see "OpenTasks" in your app drawer.

Okay, that's fine with me, thanks.

PR #203, opened 2016-02-14
clarify intent label ("Add task")

katrinleinweber:
activity_add_task_title ("Create") left this clarification up to the icon

### DIFF
--- a/opentasks/src/main/AndroidManifest.xml
+++ b/opentasks/src/main/AndroidManifest.xml
@@ -101,7 +101,7 @@
             </intent-filter>
 
             <!-- Intent filter for share intents -->
-            <intent-filter android:label="@string/activity_add_task_title">
+            <intent-filter android:label="@string/menu_add_task">
                 <action android:name="android.intent.action.SEND"/>
 
                 <category android:name="android.intent.category.DEFAULT"/>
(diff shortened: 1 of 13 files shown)


katrinleinweber:
A more extensive solution would be https://github.com/katrinleinweber/opentasks/commit/073c4dd9ba7a642e6eb6e8ecbdb1abb6c26be68e. Please let me know if that one should be in this PR instead of the above quick fix.

dmfs:
The "share as task" intent filter looks good. I'd prefer this one over the "quick fix. Thanks.

katrinleinweber:
OK, here you are :-)

dmfs:
Thanks. Looks almost good, except for the French and Brazilian translations being empty.

katrinleinweber:
Note to self: triple–check next time m-)
Just out of curiosity about Git(Hub) itself: Would it have been a problem for the PR if I had amended & force-pushed your suggested changes into the original commit?

dmfs:
I would expect that GitHub can handle the amended commit. I've never tried that though.

mkroehnert:
@katrinleinweber @dmfs pull requests are associated with a repository and a branch name.
You can push additional commits or do rebased or amended force pushes as long as the branch name stays the same.
Also see the following link: https://github.com/github/hub/issues/198

dmfs:
@mkroehnert Thanks for clarifying.

PR #231, opened 2016-04-21
Fix typo in German translation

marcoh00:
This fixes a typo in the German translation of "undo", which is "Rückgängig" instead of "Rückgänging". It's a pretty annoying one as its shown everytime you mark a task as "done" from the status bar.


saschpe:
:+1:

dmfs:
merged into staging

PR #244, opened 2016-06-01
updated spanish translation

colegota:
merged between english and old spanish versions and translated it all.


dmfs:
merged into staging

PR #247, opened 2016-06-30
Update portuguese (Portugal) translation

YSmhXQDd6Z:
I added the missing strings, but for some reason it says it cannot be merged automatically...


dmfs:
merged into staging

PR #249, opened 2016-07-04
Widget: Don't show Tasks in the future with START==DUE

korelstar:
Currently, the widget on the homescreen shows all visible and open tasks with one of the following conditions:
- Task has started (`TaskContract.Instances.INSTANCE_START + "<=" + System.currentTimeMillis()`)
- Task has no start date (`TaskContract.Instances.INSTANCE_START + " is null"`)
- Start date and due date are equal (`TaskContract.Instances.INSTANCE_START + " = " + TaskContract.Instances.INSTANCE_DUE`)

What's the motivation for the last one?

If I have a full-day Task for a day far away in the future, this task will be shown. However, the 5-minute Task which comes tomorrow will not be shown. In my view, it would be correct to hide both tasks and therefore the last condition (`TaskContract.Instances.INSTANCE_START + " = " + TaskContract.Instances.INSTANCE_DUE`) should be removed.

Have I missed something?

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/homescreen/TaskListWidgetUpdaterService.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/homescreen/TaskListWidgetUpdaterService.java
@@ -380,7 +380,7 @@ public class TaskListWidgetUpdaterService extends RemoteViewsService
 				// build selection string
 				StringBuilder selection = new StringBuilder(TaskContract.Instances.VISIBLE + ">0 and " + TaskContract.Instances.IS_CLOSED + "=0 AND ("
 					+ TaskContract.Instances.INSTANCE_START + "<=" + System.currentTimeMillis() + " OR " + TaskContract.Instances.INSTANCE_START
-					+ " is null OR " + TaskContract.Instances.INSTANCE_START + " = " + TaskContract.Instances.INSTANCE_DUE + " )");
+					+ " is null)");
 
 				if (lists != null && lists.size() > 0)
 				{


dmfs:
We added this for Apple clients which set `START== DUE` if a due date is set. Without that condition Apple users wouldn't see any tasks having a due date set.

korelstar:
Hmmm. Okay, thanks for the info. Then the current query is reasonable. However, it doesn't make sense for non-apple users. Isn't it possible to identify tasks created by Apple clients? Maybe add a new field `PRODID` (from `VCALENDAR`) to the `opentasks-provider` which then can be used in the query?

The problem is also the order of these tasks: currently, the widget have this order:
1. tasks with deadline in the past or which already began
2. tasks with `start==due` in the future
3. tasks without `due`.
It would be more useful, if 3. would come before 2. (or better: 2. is not shown at all).

If it is not possible to identify Apple tasks, we may think about an option for hiding/showing future Tasks with `START==DUE` or change the order in the widget.

dmfs:
`start==due` is clearly an edge case and strictly spoken it's not even valid as per [RFC 5545 section 3.8.2.3](https://tools.ietf.org/html/rfc5545#section-3.8.2.3), which says:

>  and the value of this property MUST be later in time than the value of the "DTSTART" property.

So we shouldn't even allow to create such a task.

What's the intention of setting `start == due` anyway? I think it's basically the same as setting no start at all, and that's essentially how we handle this (well that's how we should handle it, not sure if that's true to every extend).

What's your use case for setting `start == due`?

dmfs:
Yes, "all-day" is part of the standard. However, for an all-day task with the duration of one day you would indeed set `due = start + 1d` or `start = due - 1d`. Just like the end-date of an event, `due` is non-inclusive, which means the `due` date specifies the very first moment that the task is overdue. So if the due date is July 8th, the task should be completed by July 7th, as it's already overdue on the 8th.
RFC 5545 says about `DUE`:

>   Purpose:  This property defines the date and time that a to-do is
>    expected to be **completed**.

(emphasis is mine)
i.e. in the example above the task would be expected to be completed on the 8th, which means it must be finished during the day before.
That's also why `start == due` is not allowed, a task can't be overdue already in the very moment it starts. Or to use the example, if the task would already overdue on the 8th, you can't start on the 8th.

korelstar:
Okay, then I have misinterpreted the `due` attribute. However, I still think it's not very intuitive this way.

Nevertheless, the handling could be indeed improved by not allowing `start=due` in the app. Unfortunately, the app currently sets `start=due` by itself, if `start` or `due` with `start>due` are chosen. This strengthened my (apparently false) interpretation of the `due` attribute.

For all-day tasks, an improvement would be very easy by setting automatically (by the app) `start = due - 1d` if a `due` is set with `due<=start` (and the other way around if `start` was changed). But what is a good default interval for non-all-day tasks? Maybe 1 hour?

dmfs:
Created #262 to address the original issue of zero duration tasks.

PR #251, opened 2016-08-02
Widget: order by START date/time if DUE and PRIORITY are indistinguishable

korelstar:
Currently, the widget order is as follows:
1. by `DUE` (early due first, no due last)
2. by `PRIORITY` (high priority first, no priority last)
3. by `CREATED` (new tasks first)

In order to have a better control of tasks with the same (or no) `DUE` and `PRIORITY`, I propose to add a new condition `START` after `PRIORITY`:
1. by `DUE` (early due first, no due last)
2. by `PRIORITY` (high priority first, no priority last)
3. by `START` (early start first, no start last)
4. by `CREATED` (new tasks first)

Then, already started tasks are listed before tasks which haven't been started yet, and tasks which have been started a long time ago come before recently started tasks. This helps the user to keep the number of active tasks low in order to concentrate on finishing already started tasks. Furthermore, it allows to influence the order of tasks without `DUE` more fine-grained than using priorities.


dmfs:
merged into staging

Issue #192, opened 2016-01-16
Crash when saving a task and going to home screen


thesuperdavis:
When you save a task, the dialog box "task created" appears for about 1s. When you go to the homescreen or to the recent apps before this dialog closes, the app crashes but the task is saved.

Logcat from the crash:
>E/AndroidRuntime(22354): FATAL EXCEPTION: main
E/AndroidRuntime(22354): Process: org.dmfs.tasks, PID: 22354
E/AndroidRuntime(22354): java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
E/AndroidRuntime(22354): 	at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1365)
E/AndroidRuntime(22354): 	at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1383)
E/AndroidRuntime(22354): 	at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:636)
E/AndroidRuntime(22354): 	at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:615)
E/AndroidRuntime(22354): 	at android.support.v4.app.DialogFragment.dismissInternal(DialogFragment.java:200)
E/AndroidRuntime(22354): 	at android.support.v4.app.DialogFragment.dismiss(DialogFragment.java:166)
E/AndroidRuntime(22354): 	at org.dmfs.tasks.QuickAddDialogFragment$1.run(QuickAddDialogFragment.java:470)
E/AndroidRuntime(22354): 	at android.os.Handler.handleCallback(Handler.java:739)
E/AndroidRuntime(22354): 	at android.os.Handler.dispatchMessage(Handler.java:95)
E/AndroidRuntime(22354): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime(22354): 	at android.app.ActivityThread.main(ActivityThread.java:5310)
E/AndroidRuntime(22354): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(22354): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime(22354): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:901)
E/AndroidRuntime(22354): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:696)
[...]

PR #255, opened 2016-08-17
Fix bug #192: crash when home button pressed during quick add task sa…

lemonboston:
…ved message displayed to user. (Solution from http://stackoverflow.com/a/21701607/4247460)

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/QuickAddDialogFragment.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/QuickAddDialogFragment.java
@@ -63,7 +63,7 @@ import android.widget.TextView.OnEditorActionListener;
 /**
  * A quick add dialog. It allows the user to enter a new task without having to deal with the full blown editor interface. At present it support task with a
  * title only, but there is an option to fire up the full editor.
- * 
+ *
  * @author Marten Gajda <marten@dmfs.org>
  */
 public class QuickAddDialogFragment extends SupportDialogFragment implements OnEditorActionListener, LoaderManager.LoaderCallbacks<Cursor>,
@@ -130,6 +130,8 @@ public class QuickAddDialogFragment extends SupportDialogFragment implements OnE
 	@Retain(permanent = true, key = "quick_add_save_count", classNS = "")
 	private int mSaveCounter = 0;
 
+	private boolean mClosing;
+
 	private View mColorBackground;
 	private Spinner mListSpinner;
 
@@ -147,7 +149,7 @@ public class QuickAddDialogFragment extends SupportDialogFragment implements OnE
 
 	/**
 	 * Create a {@link QuickAddDialogFragment} with the given title and initial text value.
-	 * 
+	 *
 	 * @param titleId
 	 *            The resource id of the title.
 	 * @param initalText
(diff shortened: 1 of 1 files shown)


dmfs:
Sorry, that doesn't look like a solution, but more like a hack.
From the stack trace I'd say that the issue caused by a fragment transaction that's started after the activity has been stopped or suspended. A fix makes sure there is no further transaction started after onSaveInstanceState was called.

lemonboston:
Yes, I debugged it and it happens when QuickAddDialogFragment is dismissed automatically with delay after the 'Saved' notification was shown to user, and when in the meantime the app was put to background (home button pressed), so the activity had stopped.

(Using dismissAllowingStateLoss() (which uses FragmentTransaction.commitAllowingStateLoss()) for this case is suggested here as well:
http://stackoverflow.com/a/10275071/4247460)

At first I thought the possible state loss with commitAllowingStateLoss() is only about the state of the fragment, which would be okay for a removed fragment, because it won’t be restored anyway. But I see in the docs now that it might loose the actual commit itself as well..  
So you are probably right that its usage should be avoided then, I will look into this further.

lemonboston:
After looking into this more, I think this bug may tell that a different approach might be needed for this delayed dialog updates:
It could be argued that a Fragment transaction (as a DialogFragment dismiss) is a major UI state update which shouldn't happen while the app is in background, so any delayed transaction should be avoided because user can put it to background any time.
This is a very good post on the topic, the section "Avoid performing transactions inside asynchronous callback methods" refers to this generally:
http://www.androiddesignpatterns.com/2013/08/fragment-transaction-commit-state-loss.html
There is also a link inside from Dianne Hackborn on a similar issue:
https://groups.google.com/forum/#!msg/android-developers/dXZZjhRjkMk/QybqCW5ukDwJ

I see the following options now:
1. When "SAVE" pressed, instantly dismiss and use just a Toast for the message.
Con: There is a subtle bug with "SAVE AND CONTINUE" as well, if app is put to background after that, the keyboard comes up after the delay on the home screen. It wouldn't solve that (although for that we may only need to remove mEditText.requestFocus(); Also the 2 kind of saves would have different UX.
Pro: Quick save is quicker

2. Introduce boolean state flags and either dismiss dialog in onPause() if it is in 'closing' state, or dismiss after onResume() if it was in 'closing' state (with delay or not).
Cons:
Maybe not guaranteed to work pre-Honeycomb (see the table at the post).
More complexity, feels hacky, can be error prone if this approach is continued. 

3. Use a bit more general solution for "pausing" fragment transaction while app is in background and resuming after resume: http://stackoverflow.com/questions/8040280/how-to-handle-handler-messages-when-activity-fragment-is-paused
(again, not sure about pre-Honeycomb)

4. Handle this "Task created" UI message somehow differently. Not sure how.
If it was a Dialog, not a DialogFragment, there would be no problem, but we need the Fragment here.
A separate Dialog for this message? Not sure..

lemonboston:
In the last commit I used Option 2., so dismiss() in onPause() if the delayed dismiss has been scheduled already.
I suppose this would be the solution with the less work and less change in code and UI to fix this bug.

According to the [post](http://www.androiddesignpatterns.com/2013/08/fragment-transaction-commit-state-loss.html) there is still a chance in pre-Honeycomb to loose state but it won't throw exception.
I suppose that's fine.

lemonboston:
@dmfs Please review.
The solution is to dismiss() the DialogFragment in onPause() if the delayed dismiss() was already scheduled. (A new boolean state flag is used for that.)

dmfs:
Looks good. Merged into staging.

Issue #149, opened 2015-07-24
Click on location does not open map.


jkufner:
When clicking on location field in task detail screen, nothing happens. Calendar apps open map when clicking on event location. Tasks should do the same.

dmfs:
Agreed. We'll check how that works.

jkufner:
Thank you!

PR #270, opened 2016-09-19
PR for #149 location opens map

lemonboston:
Linking to #149.

@dmfs please review.

I haven't tested with `XmlModel`, only with local tasks and `DefaultModel`. Should I test with `XmlModel` too, and if yes, could you please advise how can I do that, where to create account, etc?


opentasks/src/main/java/org/dmfs/tasks/widget/LocationFieldView.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+        super(context, attrs, defStyle);
+        init();
+    }
+
+
+    private void init()
+    {
+        mMapOpener = new MapOpener(getContext());
+        mGestureDetector = new GestureDetector(getContext(), new GestureDetector.SimpleOnGestureListener()
```
dmfs: Can't you just make the text view or the LocationFieldView clickable?
Why the GestureDetector?

opentasks/src/main/java/org/dmfs/tasks/widget/TextFieldView.java
```diff
@@ -106,4 +106,9 @@ public void onContentChanged(ContentSet contentSet)
 			}
 		}
 	}
+
+	public String getText()
+	{
+		return mText.getText().toString();
```
dmfs: Don't use UI components for business logic. UI is either output or input, you're using it like a variable.

opentasks/src/main/java/org/dmfs/tasks/utils/MapOpener.java
```diff
@@ -0,0 +1,30 @@
...
+import android.content.Intent;
+import android.net.Uri;
+
+
+/**
+ * Handles opening map application on the device with a location query string.
+ *
+ * @author Gabor Keszthelyi
+ */
+public class MapOpener
```
dmfs: This class doesn't feel right and probably shouldn't exist. I suggest to make this a private method of the LocationFieldView.

lemonboston: Moved it to private method.

opentasks/src/main/java/org/dmfs/tasks/widget/TextFieldView.java
lemonboston: Ok, I see, that felt a little bit wrong to me too, but on the other hand I thought user would probably want to search for location that is displayed, whatever that is.
But I changed `TextFieldView` now to save the actual text in a field and use that for the getter. Please check if it's good like that.

opentasks/src/main/java/org/dmfs/tasks/widget/LocationFieldView.java
lemonboston: I spent about 2 hours trying to make the clickable thing work and I couldn't. There is of course a chance that I missed something, so let me know if I should check again and spend like 1 hour more with it.
(I tried with `clickable`, `focusable` combinations on parent and child views and some other suggestion I found on StackOverflow.)

lemonboston:
@dmfs please check comments above, otherwise it's ready for review again from my side

dmfs:
This just crashed in my emulator:
```
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: org.dmfs.tasks, PID: 2159
                  android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=geo:0,0?q=New York }
                      at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:1765)
```

dmfs:
If there is no app to open the `geo` URI scheme, we should fall back to opening Google Maps in a browser.

lemonboston:
Yeah, that's a mistake, I should have checked if the intent could be resolved..   fixed it. And it falls back to opening `http://maps.google.com/?q=<location>` now.

@dmfs Ready for review again.

opentasks/src/main/java/org/dmfs/tasks/widget/LocationFieldView.java
dmfs: The problem is that you inherit from `TextFieldView`. `TextFieldView` "linkifies" the `TextView` which consumes any clicks instead of passing them to the underlying `LinearLayout`.

I suggest to not subclass `TextFieldView`. Either create an `AbstractTextView` that both classes inherit from or just duplicate the methods in `TextFieldView`. Remember that subclassing non-abstract classes is evil ;-) (yeah, I know that we probably still do that in other places of this code base).

In this particular case I'd actually prefer the code duplication, because we're dealing with a lot of mutable dependencies (the UI) and I'd like to avoid any side effects like this one.

dmfs:
Please remove the `GestureDetector` (as per my inline comment). Also, squash all commits into a single one and submit a new PR (you'll probably have to push into a new branch).

opentasks/src/main/java/org/dmfs/tasks/widget/LocationFieldView.java
lemonboston: Ok, thanks for the tip, the culprit was `android:textIsSelectable="true"` in the style of the TextViews.
Setting that to false allows clicking on the LinearLayout now.
LocationFieldView is no not the child of TextFieldView in the new PR.

lemonboston:
Made the changes and squashed the commits, new PR is #285.
Closing this one.

Issue #220, opened 2016-03-02
Check overflow menu handling in ViewTaskFragment.


dmfs:
Apparently it's possible to delete a task in 2 pane mode even though there is no task. Otherwise #211 would be impossible (unless that is caused by a race condition).

lemonboston:
Yes, I could reproduce it:
In 2 pane mode delete a task from the bottom of a list, the detail view keeps the deleted task title and the action bar.
Any action from the action bar (complete, pin, edit, delete) result in crash (NPE for ContentSet).

lemonboston:
This cause by incorrect handling of ViewTaskFragment.loadUri(null) which is called when the detail view should display empty content.
Both the menu part and the title part are not handled correctly.
Menu could be fixed easily, the title is not that obvious.

Your (@dmfs) idea on this:

> I consider setting values to null to clear any output being bad design. Why not just hide the container or the entire fragment if there is no content to show? Or even better, replace the entire Fragment by an "EmptyDetailsFragment". The "EmptyDetailsFragment" could show some hint like "No task selected" or if there are no tasks at all replace it by a "NoTasksFragment" that says something like "Relax and sit back, there are no tasks. Tap here to add a task.".
> We probably can use the same message if there are no visible tasks (because all have been completed). We should work out the details in which case, which message should be shown.

This would be clearly better in my opinion as well.
Should we create a separate enhancement issue for this?

dmfs:
Yes, please open a new issue for that.

lemonboston:
#259 has been opened for the empty task detail fragment. When that is completed, this can be closed as well.

PR #274, opened 2016-09-21
Create EmptyTaskFragment and use it from TaskListActivity whenever no…

### LINKED ISSUE #259: Create empty task detail view (fragment)
Empty task details is displayed in the two-pane mode (on tablet) when no task is selected, or sometimes after the currently selected task is deleted, or when all tasks have been completed. Currently the same `ViewTaskFragment` is used for this as for displaying an actual task and it has bugs, see #220. As discussed there, it is better to create a separate, dedicated Fragment to display empty (or no) task.

Copied from #220:

> [...] replace the entire Fragment by an "EmptyDetailsFragment". The "EmptyDetailsFragment" could show some hint like "No task selected" or if there are no tasks at all replace it by a "NoTasksFragment" that says something like "Relax and sit back, there are no tasks. Tap here to add a task.".
> We probably can use the same message if there are no visible tasks (because all have been completed). We should work out the details in which case, which message should be shown.

Details can be discussed and worked out when we start to implement it.

lemonboston:
… task detail should be displayed.

Resolves #259 

@dmfs 
With these changes when the task at the bottom of a list is deleted, the empty fragment is displayed, so the bug #220 will be resolved with this.

However, the behavior may still not totally what user might expect (I guess we could create separate issue for these, if they are important):
- when the bottom task is deleted, empty task is displayed instead of jumping to the one above it
- when the bottom task is completed by swiping, the same task's details are shown (correctly updates as done)


dmfs:
With this version I'm getting the following exception when I'm trying to add a task on a tablet (Android 5.1) using the regular editor

```
E/AndroidRuntime( 2897): FATAL EXCEPTION: main
E/AndroidRuntime( 2897): Process: org.dmfs.tasks, PID: 2897
E/AndroidRuntime( 2897): java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=2924, result=-1, data=Intent { dat=content://org.dmfs.tasks/tasks/2 }} to activity {org.dmfs.tasks/org.dmfs.tasks.TaskListActivity}: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
E/AndroidRuntime( 2897): 	at android.app.ActivityThread.deliverResults(ActivityThread.java:3574)
E/AndroidRuntime( 2897): 	at android.app.ActivityThread.handleSendResult(ActivityThread.java:3617)
E/AndroidRuntime( 2897): 	at android.app.ActivityThread.access$1300(ActivityThread.java:151)
E/AndroidRuntime( 2897): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1352)
E/AndroidRuntime( 2897): 	at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime( 2897): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime( 2897): 	at android.app.ActivityThread.main(ActivityThread.java:5254)
E/AndroidRuntime( 2897): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 2897): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime( 2897): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
E/AndroidRuntime( 2897): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
E/AndroidRuntime( 2897): Caused by: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
E/AndroidRuntime( 2897): 	at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1489)
E/AndroidRuntime( 2897): 	at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1507)
E/AndroidRuntime( 2897): 	at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:634)
E/AndroidRuntime( 2897): 	at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:613)
E/AndroidRuntime( 2897): 	at org.dmfs.tasks.TaskListActivity.replaceDetailFragment(TaskListActivity.java:454)
E/AndroidRuntime( 2897): 	at org.dmfs.tasks.TaskListActivity.loadTaskDetailFragment(TaskListActivity.java:446)
E/AndroidRuntime( 2897): 	at org.dmfs.tasks.TaskListActivity.onItemSelected(TaskListActivity.java:409)
E/AndroidRuntime( 2897): 	at org.dmfs.tasks.TaskListActivity.onActivityResult(TaskListActivity.java:549)
E/AndroidRuntime( 2897): 	at android.app.Activity.dispatchActivityResult(Activity.java:6192)
E/AndroidRuntime( 2897): 	at android.app.ActivityThread.deliverResults(ActivityThread.java:3570)
E/AndroidRuntime( 2897): 	... 10 more
W/ActivityManager( 1522):   Force finishing activity 1 org.dmfs.tasks/.TaskListActivity
```

I don't see this with the current staging branch.

lemonboston:
Oh, it's good that you caught that, thanks.
It was caused by commiting `FragmentTransaction` (switching the empty fragment to the real one) initiated by `onActivityResult()`, before `onResume()`.
It would have been easy (although not nice) to fix this by postponing the transaction to `onPostResume()`, but I noticed that while it fixes this crash, not the correct new task is displayed in `TaskListActivity` after adding a task.
This was not correct on staging either, so after adding a new task with editor, the list keeps the selectedItemPosition and displays that task.
So I though we should fix that too, so I started looking into these loading mechanisms partly in `TaskListFragment`. I spent multiple hours with it, logging a lot. I think it is really hard to follow what's happening there and messy, for example there are some code which is executed for each child view which should be only called once probably. (I will get back to this at the end.)
But fortunately I noticed that there is this flow of opening the app from the widget and showing the selected task, which actually works. So I ended up just reusing that, calling `onNewIntent()` in `onActivityResult()`.

Still not everything is perfect, for example two subtle bugs:

**Bug#1**, steps (tablet):
1. open a task from widget (on staging as well) OR create a new task with FAB (with this branch)
2. select another task from the list (not a neighbor), delete it
3. actual: selection jumps back to task in step 1.
expected: selection jumps to neighboring task

This is because that field for this special selection is not cleared, but if it is cleared after usage than it doesn't work at all (relates to the messiness of what happens during initializing the screen, in particular `TaskListActivity.getSelectedTaskUri()` is called multiple times).

**Bug#2**, steps (tablet):
1. have a list with one task
2. edit it, set to done, save
3. actual: task details are still displayed, even though it is completed
expected: empty fragment is displayed

These bugs don't seem to be critical ones, but if we get back to this area, I think it would be worth reviewing all code related to selecting which task is displayed, and then stepping back to see if it was worth redesigning it. For example one thing that I think causes problems and caused me confusion is that the currently selected task is saved at multiple fields and classes:
`TaskListActivity#mSelectedTaskUri`
`TaskListActivity#mSelectedTaskUriOnLaunch`
`TaskListFragment#mSelectedTaskUri`
`TaskListFragment#mSelectedChildPosition`
(+ `TaskListFragment#mActivatedPositionGroup`?)
(+ `TaskListFragment#mActivatedPositionChild`?)
That `mSelectedChildPosition` is also retained in SharedPreferences as well complicating things further.
Also, the uri is passed around in lot of methods, so it is difficult to follow when and where it is saved.
I think we should decide who is the owner of it, who does it belong to and save it only there and only the uri. Perhaps in TaskListFragment.
I haven't even checked what code handles the selection if the task that was selected disappears (completed or deleted, manually or by sync), but that would as well relate to this.

@dmfs it's ready for review again.
If it looks fine, please let me know if we should create those extra bugs, possibly with a link to this comment.

opentasks/src/main/res/layout/opentasks_fragment_empty_task.xml
```diff
@@ -0,0 +1,18 @@
...
+                                                 android:layout_height="match_parent"
+                                                 android:fitsSystemWindows="true">
+
+    <android.support.design.widget.AppBarLayout
+            style="@style/task_detail_appbar">
+
+        <android.support.v4.widget.Space android:layout_width="match_parent"
+                                         android:layout_height="?attr/actionBarSize"/>
+
+        <android.support.v4.widget.Space android:layout_width="match_parent"
```
dmfs: Why this second space? I'd prefer if the toolbar of the empty fragment is aligned with the toolbar of the main fragment.

opentasks/src/main/res/values/dimens.xml
```diff
@@ -9,5 +9,6 @@
     <dimen name="opentasks_two_pane_list_width">432dp</dimen>
     <dimen name="opentasks_details_pane_overlay_top">120dp</dimen>
     <dimen name="opentasks_details_pane_margin_bottom">-80dp</dimen>
-    
+    <dimen name="task_detail_toolbar_title_margin">8dp</dimen>
```
dmfs: please use `opentasks_` prefix instead of just `tasks_`

opentasks/src/main/res/values/styles.xml
```diff
@@ -270,4 +270,12 @@
         <item name="android:src">@drawable/ic_24_undo_white</item>
     </style>
 
+    <style name="task_detail_appbar">
```
dmfs: same here use `opentasks_…`

opentasks/src/main/res/layout/opentasks_fragment_empty_task.xml
```diff
@@ -0,0 +1,18 @@
...
+                                         android:layout_height="?attr/actionBarSize"
+                                         android:layout_margin="@dimen/task_detail_toolbar_title_margin"/>
+
+    </android.support.design.widget.AppBarLayout>
+
```
dmfs: Please add a content view that contains a centered (gray) text like "No task selected".

opentasks/src/main/java/org/dmfs/tasks/TaskListActivity.java
```diff
@@ -433,6 +424,37 @@ else if (forceReload)
 	}
 
 
+	private void loadTaskDetailFragment(Uri uri)
+	{
+        Fragment detailFragment = getSupportFragmentManager().findFragmentByTag(DETAIL_FRAGMENT_TAG);
+
+		if (uri == null)
```
dmfs: There is something wrong with indentation in this method.

lemonboston: Formatted it, spaces were used instead of tabs.

lemonboston:
@dmfs updated as per your comments, ready for review again.

PR #277, opened 2016-09-29
PR for 172 share task

### LINKED ISSUE #172: Provide sharing to Android Feature
Add Feature to share tasks with other Android apps (i. e. Email, Threema,  ES File Manager)

lemonboston:
@dmfs Please check questions [here](https://github.com/dmfs/opentasks/issues/172#issuecomment-250442731), so I can update code if needed.

Should we try to localize these new strings on our own, at least for the more popular, non-symbolic languages using google translate?:
```
<string name="activity_task_details_share">Share</string>
<string name="activity_task_details_share_task">Share task</string>
<string name="share_footer">Shared by OpenTasks</string>
```

(linking to issue: resolve #172)


opentasks/src/main/java/org/dmfs/tasks/ViewTaskFragment.java
```diff
@@ -594,13 +594,24 @@ else if (itemId == R.id.pin_task)
 			persistTask();
 			return true;
 		}
+		else if (itemId == R.id.share_task)
+		{
+			shareTaskDescription();
```
dmfs: Do we really need an additional method for this one-liner?

lemonboston: No, we don't. (I inlined it)

I guess it's a matter of programming style, and I hope I won't drive you crazy with it :)
This case is not a really good example, but how I am looking at it usually is that for example here onOptionsItemSelected() method should only be concerned what happens on certain button clicks, not how that happens. That's a detail that is not relevant on his level. So if I want to put it fancily, methods have sort of abstraction levels as well, and also responsibilities.
If we don't apply single responsibility on this granularity as well, then I think methods tend to grow big and hard to follow, juggling with a lot of different kind of low level stuff.
(Again this concrete example is not the best, because the one-liner and also the conditions in onOptionItemsSelected() are straightforward, so sorry if this was irrelevant.)

dmfs:
No idea why, but I'm getting an Exception when I open the details view with this version:

```
E/AndroidRuntime( 3012): FATAL EXCEPTION: Thread-228
E/AndroidRuntime( 3012): Process: org.dmfs.tasks, PID: 3012
E/AndroidRuntime( 3012): java.lang.NullPointerException: Attempt to invoke virtual method 'void android.support.v7.widget.ShareActionProvider.setShareIntent(android.content.Intent)' on a null object reference
E/AndroidRuntime( 3012): 	at org.dmfs.tasks.share.ShareIntentFactory$1.run(ShareIntentFactory.java:69)
E/AndroidRuntime( 3012): 	at java.lang.Thread.run(Thread.java:818)
W/ActivityManager( 1241):   Force finishing activity org.dmfs.tasks/.ViewTaskActivity
W/ActivityManager( 1241):   Force finishing activity org.dmfs.tasks/.TaskListActivity
```

Android 5.1 emulator.

lemonboston:
Yes, I am aware and I am working on it, and sorry I should have posted that it was not ready for review yet.
The problem is that the share intent creation needs to wait for menu, ContentSet, Model to be loaded/created and the order of finishing them can vary. But I have some idea now how to handle  such cases generally, I will update code and comment again.

lemonboston:
I've created another PR (#282), so I am closing this.
(Branched off from staging and merged the changes there to have the Hungarian translation and nice diffs on github.)

PR #280, opened 2016-10-03
Issues/224 completed timezone

dmfs:
@lemonboston please review.

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/model/TaskFieldAdapters.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/model/TaskFieldAdapters.java
@@ -124,7 +124,7 @@ public final class TaskFieldAdapters
 	/**
 	 * Adapter for the completed date of a task.
 	 */
-	public final static TimeFieldAdapter COMPLETED = new TimeFieldAdapter(Tasks.COMPLETED, null, null);
+	public final static TimeFieldAdapter COMPLETED = new TimeFieldAdapter(Tasks.COMPLETED, Tasks.TZ, null);
 
 	/**
 	 * Adapter for the time zone of a task.


lemonboston:
I think I found a little inconsistency in how the timezone of the task gets set automatically:
Steps:
Create a task with quick add dialog, open its details and then edit.
The Time Zone field is preset to local time zone.
If I set any of the 3 dates manually there, the local time zone is kept.
But if all 3 are empty and I complete the task by either setting percent to 100% or status to done, then save, then the task's time zone will be UTC.

Otherwise, as much as I could check, it works correctly for me.

dmfs:
The problem is that a task has no time zone if no start or due date is specified. The `COMPLETED` field must not have any time zone, so a task that has only a completed date doesn't have any time zone at all. I'll check again if that's an issue, but I think it's not a blocker and we're going to release it anyway.

lemonboston:
Ok, I see, and I suppose this is not part of this bug anyway.
Just one note: If I set the completed date-time with the `COMPLETED` field on edit screen of an otherwise empty task, the local time zone is kept. So it looks like it saves a time zone in this case even with no start or due date.
If completion is done with status or percent, then it will be UTC.

dmfs:
Ah, got it. That's because when you set a specific completed date, the time zone is always set as well now (that didn't happen prior to this patch). When you just complete a task (in the details view or from a notification), the completed date is set by the task provider, which doesn't set a time zone. I suppose we can also set a time zone in the task provider in that case (I just created dmfs/opentasks-provider#9). But even if we do that, this will only work on the device the task was completed on. If the task is synced to another device, the time zone will be lost (due to iCalendar restrictions).

dmfs:
So there is nothing else to do about this PR. I've merged it into staging.

PR #331, opened 2017-03-29
Run code formatter on all java files.

dmfs:


dmfs:
On a second thought, this probably doesn't make much sense until all pending PRs have been merged.

Issue #254, opened 2016-08-16
Open task after storing it


jkufner:
I found anoying usability problem.

Curent state:
1. Creaate a task
2. Hit save button
3. Task is somewhere deep in the list

Better way:
1. Create task
2. Hit save button
3. Task opens, it is possible to edit it or pin it or whatever
4. Hit back button to get to the list

So the point is to open the newly saved task, because otherwise it is difficult to modify or pin the new task.

dmfs:
Got your point. I guess that makes sense. We'll take care of it.

lemonboston:
Should this be the behavior when saving from the quick _Add task_ dialog and the full _Create_ screen as well?

lemonboston:
Submitted pull request for the _Create_ screen (editor) case.

PR #344, opened 2017-06-27
Open task after saved with editor. #254

lemonboston:
Note: the whole method appears as change in the diff because of whitespace changes.
(I am not sure at the moment how to handle the code styles in this project..)


dmfs:
@lemonboston did you test this on tablets?

lemonboston:
Yes, I checked it, it doesn't work correctly but it doesn't work correctly in current version either, not the newly created task is selected after coming back to the 2 pane screen.
There are two updates run for selecting the item, one triggered by save and one starting from onResume() and this 2nd one selects the wrong task.
I think we could create a separate bug for this.

dmfs:
@lemonboston please rebase and resolve conflicts

lemonboston:
Rebased and conflicts resolved.

Issue #345, opened 2017-06-27
Add code style settings to git


lemonboston:
`.idea/codeStyleSettings.xml` is currently ignored, so we don't have a common code style in the project.
We can add dmfs latest version but we have make sure not to use it inside existing classes and at commit time before all files are updated at once, cause it may make diffs more difficult to review.

copynpaste:
Why should we use JetBrains specific stuff when there is Editorconfig, which is supported in more IDEs?

dmfs:
@copynpaste didn't know about Editorconfig. Nice! You're right, it's probably better to use this.

dmfs:
We went without EditorConfig for now. The EditorConfig file created by Android Studio was fairly basic and didn't account for various settings we use. Also the resulting format was actually different in a few cases.
We'll probably look into this again.

PR #347, opened 2017-06-27
Add code style project settings. #345

lemonboston:


lemonboston:
Note: code style xml is from ED-SDK

lemonboston:
Added `.editorconfig` file as well (exported from IntelliJ Idea).

dmfs:
hmm... code formatting results in different results with EditorConfig enabled. It's mostly "continuation indents" which are different (4 vs. 8).
For now I'd prefer to remove the EditorConfig file.

lemonboston:
Removed `.editorconfig` and rebased on `staging`.

Issue #93, opened 2014-08-21
Crash on add task


splack:
If you hide local tasks, don't have any other remote sources then add a task and hit plus the app crashes.

lemonboston:
Is this still relevant? I checked that if no lists are selected to display and I add a task with the plus FAB on home screen, then the task editor screen comes up where my local list is already selected, so there is no problem in this case.

dmfs:
@lemonboston try not just to hide the lists, but also to disable sync for all list. The issue occurred because the task editor only shows synced lists and it crashed when there is no synced list at all. We should catch the case that there are no writable & synced lists and show a notification in that case.

lemonboston:
@dmfs Would it be an sufficient solution to check whether the `Cursor` for the task lists is empty in `onLoadFinished` in `EditTaskFragment` (https://github.com/dmfs/opentasks/blob/a764d2d35f41b705984fa208782165d35403f951/opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java#L479),
and if it is, show a dialog message with something like:
"Please select at least one Task list in "Displayed Lists" to be able to add new tasks"
with an OK button that finishes `EditTaskActivity`.

(I've checked that the `Cursor` is empty there in this scenario)

lemonboston:
@dmfs I've just found this note from you regarding this bug:
> In time, we will support read only task lists, so there will be the case that no writable task lists are available. Now, just show a toast or snackbar if that is the case.

So I will implement it with a SnackBar or Toast and finishing the Activity as mentioned above.
(Simpler than dialog which would need to be styled, I suppose.)
Ideally we would check for this before opening the editor but I assume this is an edge case anyway so I suppose it's okay to flash it up and remove once loader completed.

dmfs:
Yes, this is an edge case. No need to put in too much effort for now. We can improve this later on.

PR #350, opened 2017-06-28
Fix crash when saving task in editor with no list. #93

lemonboston:
Note: I've reused an existing string for the toast message: "No task list available".

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
@@ -53,6 +53,7 @@ import android.graphics.drawable.ColorDrawable;
 import android.net.Uri;
 import android.os.Build.VERSION;
 import android.os.Bundle;
+import android.support.v4.app.FragmentActivity;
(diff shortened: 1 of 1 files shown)


opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
```diff
@@ -476,7 +477,13 @@ public void onSaveInstanceState(Bundle outState)
...
 	public void onLoadFinished(Loader<Cursor> loader, Cursor cursor)
 	{
-		mTaskListAdapter.changeCursor(cursor);
+		if (cursor == null || cursor.getCount() == 0)
+		{
+			showNoListMessageAndFinish();
+			return;
+		}
+
+        mTaskListAdapter.changeCursor(cursor);
```
dmfs: code style?!

lemonboston: Do you mean the guard with the `return`? To change that?
Or the code format style? I've applied that to these lines... but also rebased to `staging` in the meantime and I don't see those big indents now..   Can you please check again now?

PR #355, opened 2017-07-03
Share task details with other apps. #172

lemonboston:
(continuing https://github.com/dmfs/opentasks/pull/282 , cause that was closed accidentally when branch was removed)

TODO German translations
(https://github.com/dmfs/opentasks/issues/172#issuecomment-312687018)

PR #282, opened 2016-10-04
PR for #172 share task merge

lemonboston:
fixes #172 

@dmfs please review

I've realized that the words "Share" and "Send" are both commonly used for this feature. Should we change texts to "Send", "Send task", "Sent by OpenTasks"?
I've used this _send_ versions in the Hungarian translation, it felt a little bit better, but I am not sure.

TODO German translations

PR #368, opened 2017-07-05
Share task details with other apps. #172

lemonboston:
(Continuation of https://github.com/dmfs/opentasks/pull/282)

TODOs
- [x] add German transaltions of new text: https://github.com/dmfs/opentasks/issues/172#issuecomment-312724705
- [x] use template engine for creating the shared text
https://github.com/dmfs/opentasks/pull/355#issuecomment-312869908
https://codeka.github.io/carrot/
-> create separate task https://github.com/dmfs/opentasks/issues/372


lemonboston:
The `carrot` library has Java8 codebase (lambda, Java8 JDK classes and methods), so we can't use that.
This is blocked until we decide what solution to use.

dmfs:
Ok, please open another story for the template engine. I'd be willing to contribute a patch to carrot to make it Android compliant (see codeka/carrot#1). After all it's probably not that much work.
For the time being, please rebase on master and squash.

lemonboston:
Story created for the template engine: https://github.com/dmfs/opentasks/issues/372

Squashed and rebased, ready for review.

opentasks/src/main/java/org/dmfs/tasks/utils/TaskText.java
```diff
@@ -0,0 +1,28 @@
...
+package org.dmfs.tasks.utils;
+
+/**
+ * A {@link CharSequence} that describes one or more property of a task in some format.
+ * <p>
+ * (E.g.: The text for sharing a task details for other apps.)
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface TaskText extends CharSequence
```
dmfs: This interface looks a bit like over-engineering to me, especially since you never use it for a parameter or variable.

I suggest to get rid of it.

opentasks/src/main/java/org/dmfs/tasks/share/ShareTaskText.java
```diff
@@ -0,0 +1,278 @@
...
+public class ShareTaskText extends AbstractStringCharSequence implements TaskText
+{
+
+    public ShareTaskText(ContentSet contentSet, Model model, Context context)
+    {
+        super(new Composer(contentSet, model, context).create());
+    }
+
+
+    private static class Composer
```
dmfs: This class doesn't make much sense. You could just create a private method `String create(ContentSet contentSet, Model model, Context context)` which does the job.

opentasks/src/main/java/org/dmfs/tasks/utils/AbstractStringCharSequence.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.dmfs.tasks.utils;
+
+/**
+ * Base class for {@link CharSequence}s that use a {@link String} delegate value.
+ *
+ * @author Gabor Keszthelyi
+ */
+public abstract class AbstractStringCharSequence implements CharSequence
```
dmfs: This class is a bit weird. It's essentially an adapter for String to CharSequence, which doesn't make much sense, because Strings already are CharSequences.
I'll accept this for now, but please create a story to replace this by something like `Lazy` which seems to be the intent behind this.

lemonboston:
Moved this back to "In Progress" reassigned to myself to attend to it later.

opentasks/src/main/java/org/dmfs/tasks/utils/TaskText.java
lemonboston: Right, I removed it.

opentasks/src/main/java/org/dmfs/tasks/share/ShareTaskText.java
lemonboston: I created it to avoid passing around those parameters to all methods. But I've removed it now.
Just to realize that it may be good as a Factory...  Anyway, added TODO as well to replace with Factory when we have those from the new library.

opentasks/src/main/java/org/dmfs/tasks/utils/AbstractStringCharSequence.java
lemonboston: True, added TODO there to replace it when those types are available from the planned java tools library. (Will add a ticket when github is responsive)

lemonboston:
Ready for check again.

lemonboston:
Squashed and rebased onto master.

opentasks/src/main/java/org/dmfs/tasks/share/ShareTaskText.java
```diff
@@ -0,0 +1,108 @@
...
+ Shared by OpenTasks
+ */
+
+
+/**
+ * {@link CharSequence} for sharing task information, uses <code>carrot</code> template engine.
+ *
+ * @author Gabor Keszthelyi
+ */
+public class ShareTaskText extends AbstractCharSequence
```
dmfs: class not final nor abstract

opentasks/src/main/java/org/dmfs/tasks/share/TaskBindings.java
```diff
@@ -0,0 +1,136 @@
...
+
+
+    @Override
+    public boolean isEmpty()
+    {
+        return mDelegate.isEmpty();
+    }
+
+
+    private static final class SuffixBindings implements Bindings
```
dmfs: This should be the `TaskBindings` class. The `SingletonBindings` of your `TaskBindings` should be created in `ShareTasktext.OutputFactory.create()` (just like you bind the tformat function).

opentasks/src/main/java/org/dmfs/tasks/share/TitleText.java
```diff
@@ -0,0 +1,36 @@
...
+import org.dmfs.tasks.model.TaskFieldAdapters;
+import org.dmfs.tasks.utils.AbstractStringCharSequence;
+
+
+/**
+ * {@link TaskText} for simply the title.
+ *
+ * @author Gabor Keszthelyi
+ */
+public class TitleText extends AbstractStringCharSequence implements CharSequence
```
dmfs: Class not final nor abstract.

opentasks/src/main/java/org/dmfs/tasks/utils/AbstractStringCharSequence.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+// TODO Remove this class when the Factory, Lazy, AbstractCachingCharSequence line is available in java tools library
+public abstract class AbstractStringCharSequence implements CharSequence
+{
+    private final String mValue;
+
+
+    public AbstractStringCharSequence(String value)
```
dmfs: Why does it take a String? Couldn't it just take any other `CharSequence`?

dmfs: Also, can this be merged with `AbstractCharSequence`

opentasks/src/main/res/raw/sharetask.template
```diff
@@ -0,0 +1,18 @@
+{{ $task.title }}
+============
+{% if $task.description %}
+{{ $task.description }}
+{% end %}{% if $task.checklist %}{% for item in $task.checklist %}
+[{% if item.checked %}X{% else %} {% end %}] {{ item.text }}{% end %}
```
dmfs: Ẁith the latest version of carrot you can replace
```
[{% if item.checked %}X{% else %} {% end %}]
```
with
```
{{ item.checked and "X" or " " }}
```
which is easier to read and does the same otherwise.
This works because boolean operators now return the original value which decided on the outcome of the operation.

opentasks/src/main/res/raw/sharetask.template
```diff
@@ -0,0 +1,18 @@
+{{ $task.title }}
+============
+{% if $task.description %}
```
dmfs:
```
{% if $task.description %}
{{ $task.description }}
{% end %}
```
Can be replaced with
```
{{ $task.description or "" }}
```

opentasks/src/main/res/values/strings.xml
```diff
@@ -247,7 +249,10 @@
     <string name="accept">Accept</string>
     <string name="now">now</string>
 
-    <!-- Sharing intent -->
+    <!-- Incoming share -->
     <string name="share_as_task">Create Task</string>
 
+    <!-- Outgoing share -->
+    <string name="opentasks_share_footer">Shared by OpenTasks</string>
```
dmfs: I'm not certain about the preposition "by" in this case.
If I use OpenTasks to share a task with you, it has been "shared **by** me **with** you **xxx** OpenTasks". So what's the right preposition here?

proguard.cfg
```diff
@@ -1,83 +0,0 @@
--optimizationpasses 5
```
dmfs: Why did you delete the proguard file altogether? Please keep it.

dmfs:
btw, please also rebase

opentasks/src/main/java/org/dmfs/tasks/share/ShareTaskText.java
lemonboston: done

opentasks/src/main/java/org/dmfs/tasks/share/TaskBindings.java
lemonboston: Right, I guess it's better if prefix is defined independently at the place of usage. Updated.

opentasks/src/main/java/org/dmfs/tasks/share/TitleText.java
lemonboston: updated

opentasks/src/main/java/org/dmfs/tasks/utils/AbstractStringCharSequence.java
lemonboston: These were in part leftover from earlier and I didn't want to spend time on them since we would have support for this from jems, but I've reworked them now.
There is `DelegatingCharSequence`, and `CachingCharSequence` which uses `OnDemand` and `Cached`. I chose `OnDemand` here, but let me know if you already have another name for that.

opentasks/src/main/res/raw/sharetask.template
lemonboston: If I replace that then I can't make the formatting right when the description is not present. Too many empty lines. So I left this untouched.

opentasks/src/main/res/raw/sharetask.template
lemonboston: Wow, quite advanced evaluation syntax supported now. I've update it.

opentasks/src/main/res/values/strings.xml
lemonboston: Maybe "from" ?  The other I thought about is "via", but I think "from" could be straightforward. "Shared from OpenTasks". What do you think?

proguard.cfg
lemonboston: There were 2 instances of this file with only a small difference in formatting, so identical in content.
One under the root directory and one under the app module, so I removed the one in the root, it looked like that it was not used. `build.gradle` in app module references the one next to it.

opentasks/src/main/res/raw/sharetask.template
dmfs: Are you sure? How would this issue in any blank lines if there is no description?
```
{{ $task.title }}
============
{{ $task.description or "" }}{% if $task.checklist %}{% for item in $task.checklist %}
[{% if item.checked %}X{% else %} {% end %}] {{ item.text }}{% end %}
{% end %}{% if $task.location %}
{{ $R.string.task_location}}: {{ $task.location }}{% end %}{% if $task.start %}
{{ $R.string.task_start}}: {{ tformat.apply($task.start) }}{% end %}{% if $task.due %}
{{ $R.string.task_due}}: {{ tformat.apply($task.due) }}{% end %}{% if $task.completed %}
{{ $R.string.task_completed}}: {{ tformat.apply($task.completed) }}{% end %}{% if $task.priority %}
{{ $R.string.task_priority}}: {{ $task.priority }}{% end %}{% if $task.privacy %}
{{ $R.string.task_classification}}: {{ $task.privacy }}{% end %}{% if $task.status %}
{{ $R.string.task_status}}: {{ $task.status }}{% end %}{% if $task.url %}
{{ $task.url }}{% end %}

--
{{ $R.string.opentasks_share_footer}}
```

This should result in exactly the same text.

proguard.cfg
dmfs: ah, right.

lemonboston:
Rebased to master, ready for review again.

opentasks/src/main/res/raw/sharetask.template
dmfs: I've just tested this and it works for me.

opentasks/src/main/res/values/strings.xml
dmfs: "via" would refer to the transport like in "via email" or "via SMS".
I've though about "from" as well, but it sounds strange to me. I'll try to ask a native speaker.

opentasks/src/main/res/raw/sharetask.template
lemonboston: I've just tried and it produces different format that the original plan was, there is no empty line after the title, and also there is no empty line after the description, for me.

Wasn't this the plan?:
With description + checklist:
```
 <task title>
 ============

 <task description>

 [X] checked list item
 [ ] unchecked list item

 Location: <location>
 Start: <start date time> <timezone>
 Due: <due date time> <timezone>
 Completed: <due date time> <timezone>
 Priority: <priority text>
 Privacy: <privacy text>
 Status: <status text>
 <url>

 --
 Shared by OpenTasks
```
With checklist:
```
 <task title>
 ============

 [X] checked list item
 [ ] unchecked list item

 Location: <location>
 Start: <start date time> <timezone>
 Due: <due date time> <timezone>
 Completed: <due date time> <timezone>
 Priority: <priority text>
 Privacy: <privacy text>
 Status: <status text>
 <url>

 --
 Shared by OpenTasks
```
Without them:
```
 <task title>
 ============

 Location: <location>
 Start: <start date time> <timezone>
 Due: <due date time> <timezone>
 Completed: <due date time> <timezone>
 Priority: <priority text>
 Privacy: <privacy text>
 Status: <status text>
 <url>

 --
 Shared by OpenTasks
```
So to keep an empty lines always below the title.

lemonboston:
Please note the extra commit: I've renamed `CachingCharSequence` to `LazyCharSequence` and corrected the template comment in `SharedTaskText`.

opentasks/src/main/res/raw/sharetask.template
dmfs: ah, right. Ok keep it like this then.

opentasks/src/main/res/values/strings.xml
dmfs: According to https://ell.stackexchange.com/questions/141931/preposition-for-shared "via" is a valid option. So let's go with that.

lemonboston: Updated to `Shared via OpenTasks`.

dmfs:
@lemonboston please rebase

lemonboston:
Squashed and rebased, and created ticket to replace some classes: https://github.com/dmfs/opentasks/issues/436

Issue #354, opened 2017-07-03
Move TaskContract to separate module


dmfs:
At present the TaskContract is declared in opentasks-provider. To import the contract in another project you need to pull in the entire provider or copy the file.
We should move the TaskContract to a separate module called `opentasks-contract` which is then imported by the provider.

PR #377, opened 2017-07-10
Move TaskContract to separte module. #354

lemonboston:
Note:
`TaskContract` is now in a new module, along with now top level class `UriFactory`.
`TaskContract.taskAuthority(context)` has been moved to new `AuthorityUtil` in `provider` module since it uses the class `TaskProvider`.
I saw the TODO there to create an `Authority` class instead of the static method, I suppose we can do that as a separate task.


dmfs:
Also, I think `AuthorityUtil` should go to the `opentasks` module, because that's the only place where it's being used and where it makes sense to use it.

lemonboston:
Rebased on `master`.

Tried to move `AuthorityUtil` to `opentasks` app module but it is currently in use in `providers` module `ContentOperation.fire()` method as well, so I can't move unless doing other changes as well.

lemonboston:
Rebased to master and fixed conflicts.

Issue #371, opened 2017-07-06
Fix Lint errors


lemonboston:
There are 18 errors reported by Lint currently, we can probably fix them.

(Lint report attached.)
[lint-results.txt](https://github.com/dmfs/opentasks/files/1128816/lint-results.txt)

@dmfs Please prioritize this (move to Backlog if you see fit).

PR #378, opened 2017-07-10
Fix lint errors. #371

lemonboston:
Notes:

---

I've removed these:
```
    <string name="notification_task_due_title">%1s</string>
    <string name="notification_task_start_title">%1s</string>
```
and the just the title directly for the notification, since this resource was the same in all langauges except in NL which was probably incorrect.

---

There was a missing, commented out super call in `TextFieldEditor` `onFinishInlfate()`, do you know if there was any reason for that or okay to put it in?

---

Added a CZ translation here ("few"):
```
    <plurals name="number_of_tasks">
        <item quantity="one">%d úkol</item>
        <item quantity="few">%d úkoly</item>
        <item quantity="other">%d úkolů</item>
    </plurals>
```
Lint said "few" is required for CZ, so I checked with Google Translate and it seemed to understand the plurals well, it was different for "1 task", "2 tasks", "200 tasks" with the first and last version corresponding to our existing translation, so I think that "few" should be good.

---

I've also corrected 2 translations in Hungarian.

---

In `TaskProvider`, I've removed the `PackageManager.GET_PROVIDERS` flag from here:
```
packageManager.getProviderInfo(new ComponentName(context, providerClass), PackageManager.GET_PROVIDERS | PackageManager.GET_META_DATA);
```
It is a not allowed flag there, I think it is supposed to be used with other methods.

---

Added this permission to `providers` module, since it was required by a method call in the module:
```
<uses-permission android:name="android.permission.GET_ACCOUNTS"/>
```

### DIFF
--- a/opentasks-provider/src/main/AndroidManifest.xml
+++ b/opentasks-provider/src/main/AndroidManifest.xml
@@ -14,6 +14,8 @@
             android:permissionGroup="android.permission-group.PERSONAL_INFO"
             android:protectionLevel="dangerous"/>
 
+    <uses-permission android:name="android.permission.GET_ACCOUNTS"/>
+
     <application>
         <provider
                 android:name="org.dmfs.provider.tasks.TaskProvider"
--- a/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
+++ b/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
@@ -1345,7 +1345,7 @@ public final class TaskProvider extends SQLiteContentProvider implements OnAccou
         // On Android 2.3+ we just call the appropriate method
         try
         {
-            return packageManager.getProviderInfo(new ComponentName(context, providerClass), PackageManager.GET_PROVIDERS | PackageManager.GET_META_DATA);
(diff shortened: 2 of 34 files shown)


dmfs:
Technically we still support SDK level 9, did you check what happens after you've added all the version checks to the list view?

lemonboston:
I didn't check it on an <API11 emulator, mostly because all the code that has been surrounded newly by version checks are already conditional and look like to be about only selected item state on tablet, I suppose. So I thought that it cannot be worse than calling into a method that doesn't exist.

Now, I've installed the emulators (phone and tablet, API 10 (I didn't see 9 in the list)), found an unrelated bug: https://github.com/dmfs/opentasks/issues/380 (pls prioritize if needed)
I checked to select tasks in the list on tablet, 2 pane mode, whether it seems to work or not, and I didn't see difference in current `master` and this branch.

lemonboston:
And sorry, I had missed to push my last changes when I submitted the PR.. I pushed them now, they were already squashed, so you cannot see it as separate commit. (I also rebased on `master` now)
The last change was about changing `Due in %s` to `Due in %d` in string resource for the year and using `int` in code correspondingly.

Issue #380, opened 2017-07-11
Crash on API 10 (Gingerbread 2.3.3) when clicking quick add task


lemonboston:
```
java.lang.ClassCastException: android.support.v7.widget.TintContextWrapper
  at org.dmfs.tasks.groupings.ByList$2$1.onClick(ByList.java:281)
  at android.view.View.performClick(View.java:2485)
  at android.view.View$PerformClick.run(View.java:9080)
  at android.os.Handler.handleCallback(Handler.java:587)
  at android.os.Handler.dispatchMessage(Handler.java:92)
  at android.os.Looper.loop(Looper.java:130)
  at android.app.ActivityThread.main(ActivityThread.java:3683)
  at java.lang.reflect.Method.invokeNative(Native Method)
  at java.lang.reflect.Method.invoke(Method.java:507)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:839)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:597)
  at dalvik.system.NativeStart.main(Native Method)
```

PR #388, opened 2017-07-12
Replace Context to Activity casting with getSystemService() usage to …

### LINKED ISSUE #384: Google Play Crash Report: java.lang.ClassCastException
```
java.lang.ClassCastException: 
  at org.dmfs.tasks.groupings.ByPriority$2$1.onClick(ByPriority.java:268)
  at android.view.View.performClick(View.java:4640)
  at android.view.View$PerformClick.run(View.java:19425)
  at android.os.Handler.handleCallback(Handler.java:733)
  at android.os.Handler.dispatchMessage(Handler.java:95)
  at android.os.Looper.loop(Looper.java:146)
  at android.app.ActivityThread.main(ActivityThread.java:5593)
  at java.lang.reflect.Method.invokeNative(Native Method:0)
  at java.lang.reflect.Method.invoke(Method.java:515)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1283)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1099)
  at dalvik.system.NativeStart.main(Native Method:0)
```

lemonboston:
…fix ClassCastExceptions. #380

---

The problem looks to be the casting of `Context` to `Activity` when it was 'context-wrapped':
```java
QuickAddDialogFragment.newInstance(tag).show(((FragmentActivity) v.getContext()).getSupportFragmentManager(), null);
```
-->
```
java.lang.ClassCastException: android.support.v7.widget.TintContextWrapper
  at org.dmfs.tasks.groupings.ByList$2$1.onClick(ByList.java:281)
```

I used the `getSystemService()` override solution to fix this, `Activity` can provide its reference through that method.

This should fix #384 and #385 as well.

Notes for after reviewing the code:
The 2 Activities that don't extend `BaseActivity` now:
- `DashClockPreferenceActivity extends PreferenceActivity` - I assume `ContextActivity` is not used from within this Activity, so it's okay. By the way, where is this Activity used, how is it started? I don't see direct usage in code, nor intent-filter for implicit intent.
- `ColorPickerActivity  extends FragmentActivity` 'overdefined' in the manifest. No extra code added in this app, so I suppose it's also safe from `ContextActivity` point of view.


dmfs:
@lemonboston please resolve conflicts and rebase

lemonboston:
Rebased.

dmfs:
@lemonboston more conflicts showed up after merging earlier PRs. Please rebase once again.

lemonboston:
Moved this back to "In Progress" to attend to it later.

dmfs:
Ok, I've looked more closely into this and I'm wondering if this is the right approach. The problem is: This essentially creates an implicit dependency. `ByList` now implicitly requires to run inside a `BaseActivity` without any compile time error if that's not the case.

I can see two other solutions:
* pass the `Activity` to the method which needs it
* pass a pigeon cage which takes a `ContentSet` and let the fragment transaction be handled by the activity

At present I think the first option is by far the easiest one. I mean, `ByList.GROUP_VIEW_DESCRIPTOR` depends on the activity. The correct way of expressing this is to make it a constructor parameter.
That being said, the descriptor stuff is certainly subject to a refactoring.

lemonboston:
I've made the suggested changes, removed the `Activity` access through `Context.getSystemService()` and added `Activity` to `ByList` and `ByPriority` constructors. Ready for review again.

Issue #390, opened 2017-07-24
Dutch translation updates


dmfs:
New Dutch translation submitted by Toon:
```
<?xml version="1.0" encoding="utf-8"?>
<resources>

    <!-- Manifest stuff -->
    <string name="app_name">OpenTasks</string>

    <!-- Task list -->
    <string name="menu_add_task">Taak toevoegen</string>
    <string name="menu_add_task_condensed">Toevoegen</string>
    <string name="menu_show_completed">Afgeronde taken weergeven</string>
    <string name="menu_hide_completed">Afgeronde taken verbergen</string>
    <string name="menu_show_completed_condensed">Afgeronde taken weergeven</string>
    <string name="menu_sync_now">Ververs</string>
    <string name="menu_sync_now_condensed">Ververs</string>
    <string name="menu_search_hint">Zoeken...</string>
    <string name="toast_task_completed">\"%1s\" afgerond.</string>
    <string name="toast_task_uncompleted">\"%1s\" heropend.</string>
    <string name="toast_x_removed">\"%1s\" verwijderd.</string>
    <string name="task_group_title_list">Taken</string>
    <string name="task_group_title_start">Taken beginnen</string>
    <string name="task_group_title_due">Taken te doen</string>
    <string name="task_group_title_priority">Taak prioriteit</string>
    <string name="task_group_title_progress">Taak voortgang</string>
    <string name="task_group_title_default">Taken</string>
    <string name="quick_add_task_created">Taak toegevoegd</string>
    <string name="initial_local_task_list_name">Mijn taken</string>

    <plurals name="number_of_tasks">
        <item quantity="one">%d taak</item>
        <item quantity="other">%d taken</item>
    </plurals>

    <!-- Task detail view -->
    <string name="activity_task_details_edit">Wijzig</string>
    <string name="activity_task_details_edit_task">Taak wijzigen</string>
    <string name="activity_task_details_delete">Verwijder</string>
    <string name="activity_task_details_delete_task">Taak verwijderen</string>
    <string name="activity_task_details_complete">Atronden</string>
    <string name="activity_task_details_complete_task">Taak afronden</string>
    <string name="activity_task_details_pin">Vastzetten</string>

    <!-- The title of the button to add a new item to a check list. -->
    <string name="opentasks_checklist_add_item">Item toevoegen</string>

    <!--
    The hint that's shown for a check list item that has no text.
    Should be something like "Item title", "Empty item", "new item" or "Enter item title", whatever works best in your language.
[...]

PR #401, opened 2017-08-03
Update Dutch translations with file from Toon. #390

lemonboston:
I can't of course check the translation correctness but the diff looked good. 
Just replaced "..." sequences with ellipsis character as Android suggested.

### DIFF
--- a/opentasks/src/main/res/values-nl/strings.xml
+++ b/opentasks/src/main/res/values-nl/strings.xml
@@ -5,14 +5,14 @@
     <string name="app_name">OpenTasks</string>
 
     <!-- Task list -->
-    <string name="menu_add_task">Voeg taak toe</string>
-    <string name="menu_add_task_condensed">Voeg toe</string>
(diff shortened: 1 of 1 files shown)


dmfs:
you can use translate.google.com to see if a translation looks correct.
Please approve when you're done.

lemonboston:
I've gone over the changes one by one and made some changes:
- removed the dots (...) from all of the as at some places they were clearly incorrect, at other places probably questionable, but it seemed better to remove them
- many changes seemed to be similar or even same meanings according to google, I didn't touch those
- there were a few where it was clearly incorrect, translating the resource name instead of the content, like "Edit task title" in Dutch, I replaced them with the previous version
- made a few other common sense changes

I don't know the origination of this new translation but if it's from a Dutch person and not generated with some tool, I suppose we can assume the changes which result is similar English meaning are improvements. And with that and with this review, I think it is probably a better version then the previous one. (Some English, previously untranslated words have also been translated in here)

Issue #372, opened 2017-07-07
Use template engine for creating the shared task text


lemonboston:
The composition of the text when sharing a task has been implemented with 'manual' concatenation logic, it would be better to use a template engine for it.
See https://github.com/dmfs/opentasks/pull/368#issuecomment-313523823 as well.

dmfs:
@lemonboston please check if the changes in codeka/carrot#1 do the trick

lemonboston:
Yes, it works but I had to still fork it to add
```groovy
targetCompatibility = '1.7'
sourceCompatibility = '1.7'
```
because by default jitpack builds it with Java 8 and that can't be imported.

I also added a custom carrot `ResourceLocater` to look up raw Android resource:
https://github.com/dmfs/opentasks/blob/5cfddb6574b64ff41e9acfaf0a159ec3f558a077/opentasks/src/main/java/org/dmfs/tasks/share/RawIdResourceLocator.java#L37-L37
Usage in:
https://github.com/dmfs/opentasks/blob/5cfddb6574b64ff41e9acfaf0a159ec3f558a077/opentasks/src/main/java/org/dmfs/tasks/share/CarrotShareTaskText.java#L53-L53

So there is this branch 'mx/372-template-engine-for-share' with working code but they are just initial commits, I will continue with it and review changes there later.

lemonboston:
@dmfs I leave it on hold until you confirm to go ahead with implementation.
The fork is here: https://github.com/lemonboston/carrot
An option may be to add a gradle task to the source project to create a jar with Java 7, to avoid having to maintain a fork.

lemonboston:
I am not sure if it's worth keeping a fork of carrot which adds those two compatibility lines to build.gradle and import with jitpack, or it's simpler if we just build the jar by checking out the project and adding those lines.
For now I move on with latter, so I removed my fork and added the jar to the project with a note .txt about how to build it. (branch `mx/372-template-engine-for-share`).

dmfs:
We certainly won't maintain a fork, so ideally we create a PR to get our changes upstream. That however requires to "fix" the tests to work with Java 7.

PR #402, opened 2017-08-09
Share task text with carrot template engine. #372

lemonboston:
This pull request replaces the 'manual' assembly of the shared task text with using the carrot template engine.

---

@dmfs I am submitting this now but there are TODOs left listed in `ShareTaskText`, mostly related to `androidcarrot` about the issues opened [there](https://github.com/dmfs/androidcarrot/issues). Shall I create pull requests for those?


dmfs:
let's wait for the next carrot release. All our changes have been merged so we don't need to use our own fork.

lemonboston:
We don't use a fork any more here, `carrot` is imported directly with commit hash with jitpack.
But we use `AndroidBindings` from your `androidcarrot` extension library. Shall we copy the necessary files from that to OpenTasks?

The related pending TODOs are listed here:
https://github.com/dmfs/opentasks/blob/174c1871651f6578590cf4f0b3f5e8590cd2ee8e/opentasks/src/main/java/org/dmfs/tasks/share/ShareTaskText.java#L67-L67

opentasks/proguard.cfg
```diff
@@ -77,4 +77,6 @@
     java.lang.String TAG;
     @org.dmfs.android.retentionmagic.annotations.* <fields>;
     private long mId;
-}
\ No newline at end of file
+}
+
+-dontwarn au.com.codeka.carrot.**
```
dmfs: is this still necessary?

lemonboston: I was just checking that, no there were no more warnings, so I've removed it.

lemonboston:
Ready for review now.
There are those `Lazy`, `Factory` types added here which may be replaced with the final ones I assume coming at some point in `jems`.

Issue #399, opened 2017-08-02
Pinned task triggers the notification LED


mase76:
Hi!
When I pinned a task, the LED notifies me all the time.
Notification via LED should be for task alarms only.

dmfs:
Which device and Android version? I can't reproduce this on my device.

mase76:
I have a Sony Xperia X with Omnirom. Android is 7.1.2.
Notification is only when screen is off, not all the time.

lemonboston:
I can't really try to replicate this, the emulator's screen can't be turned off and also that LED on it is probably for illustration purposes.

Controlling the LED for notification would be done with [.setLight()](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.Builder.html#setLights(int,%20int,%20int)) but we don't use that in the codebase.

I tried to search for the issue for Sony Xperia, didn't find anything meaningful, searching for Omnirom and LED blinking does have some results but none of them clearly about this.

@mase76 Does this happen only with OpenTasks notifications? So if you have a notification from another app which you let on and turn off the screen, the LED doesn't blinks?

The other thing we can look into whether the notification priority can affect this, I see that we set HIGH at some places. To check whether anything has changed in Nougat for example regarding this.

lemonboston:
We also set the default lights flag here:
```
        if (withSound)
        {
            builder.setDefaults(Notification.DEFAULT_ALL);
        }
        else
        {
            builder.setDefaults(Notification.DEFAULT_LIGHTS);
        }
```
But if this worked so far then I suppose it should probably work the same way for Nougat as well.

mase76:
Other apps with notifications, for example K9mail don't notify by LED. But in the K9mail options I can enable LED notification. Openkeychain, Orbot, or any other app also do not notify by LED, when it is not wanted.

lemonboston:
@mase76 And when you enable LED notification in K9mail and have a notification from that when screen turned off, is the LED blinking the same way as for OpentTasks pinned tasks? By the way, is it blinking or continuous light?

mase76:
The LED in K9 blinks  the same way when the screen is off. The only difference is the colour. K9 blinks blue, OT blinks white.

lemonboston:
I can actually reproduce this with my device as well (Nexus 5, Android 6), I was under the wrong assumption that it is Android 7 related.
[...]

Issue #305, opened 2016-10-31
option for disabling vibrations on notification


mkae:

dmfs:
Implemented in #403

PR #403, opened 2017-08-10
Notification signal settings. #305

lemonboston:
This pull requests adds 3 extra entries to the Settings screen for customizing notifications by enabling/disabling the sound, vibration, and lights.

A new `NotificationSignal` type is introduced to handle this aspect better.
The intent extra indicating that a notification should be silent has been renamed to "no signal" along with boolean variables representing it, in order to make it less ambiguous - "silent" and especially "withSound" can refer to only sound, while as I understand in these cases we need the notification to not signal in any way.

The changes also fix https://github.com/dmfs/opentasks/issues/399 by using no signal for pinned notification.

Knows issue:
Task started notification for already pinned task doesn't blink the LED, only plays sound and vibrate (when all is enabled). I don't know why is this, maybe it's a system behavior that updating existing notification doesn't use the lights.
But I suppose it's not that important, let me know if I should look into it, create a separate task maybe.


dmfs:
@lemonboston please rebase on `settings/master` and while you're at it, rename all the strings starting with `schedjoules…` to `opentasks…` ;-)

lemonboston:
Rebased and fixed the strings (oops..)

opentasks/src/main/res/values/strings.xml
```diff
@@ -250,4 +250,15 @@
...
+    <!-- Settings -->
+    <!-- Section title in settings for the notification signals -->
+    <string name="opentasks_settings_notification_signals_section_title">Notification signals</string>
+    <!-- Notification Sound enabling/disabling entry label in settings -->
+    <string name="opentasks_settings_notification_sound">Sound</string>
+    <!-- Notification Vibrate enabling/disabling entry label in settings -->
+    <string name="opentasks_settings_notification_vibrate">Vibrate</string>
+    <!-- Notification "Pulse light" enabling/disabling entry label in settings -->
+    <string name="opentasks_settings_notification_lights">Pulse light</string>
+
```
dmfs: German translations:

Notification signals: Benachrichtungseinstellungen
Sound: Ton
Vibrate: Vibration
lights: LED

opentasks/src/main/res/xml/app_preferences.xml
```diff
@@ -1,8 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <CheckBoxPreference
-            android:key="@string/schedjoules_pref_notification_enabled"
+    <SwitchPreference
+            android:key="@string/opentasks_pref_notification_enabled"
             android:title="@string/enable_alarms"
```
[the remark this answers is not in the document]
lemonboston: Done

opentasks/src/main/java/org/dmfs/tasks/notification/signals/CustomizableSignal.java
```diff
@@ -0,0 +1,62 @@
...
+    public CustomizableSignal(boolean sound, boolean vibrate, boolean light)
+    {
+        mSound = sound;
+        mVibrate = vibrate;
+        mLight = light;
+    }
+
+
+    @Override
+    public int defaultsValue()
```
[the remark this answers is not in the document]
lemonboston: True that they can be separated more, I remember that I thought about that for a second as well.

So I've refactored it now. I didn't use the names `Acoustic`, `Haptic` `Visual` because I think it would just add a parallel terminology even if in some terms it could be more appropriate. I think it's simpler to keep in line the the constant names in `Notification` which also corresponds to settings UI names as welll. So I just used `Sound`, `Lights`, `Vibration`.

The result is this:
```
return new Lights(lights, new Vibration(vibrate, new Sound(sound, new NoSignal()))).value();
```

opentasks/src/main/java/org/dmfs/tasks/notification/signals/NotificationSignal.java
```diff
@@ -0,0 +1,36 @@
...
+ * @author Gabor Keszthelyi
+ */
+public interface NotificationSignal
+{
+    /**
+     * Returns the value that can be used for {@link NotificationCompat.Builder#setDefaults(int)}.
+     * <p>
+     * (<code>0</code> means no signal.)
+     */
+    int defaultsValue();
```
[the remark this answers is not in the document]
lemonboston: You're right, I agree that that name was not good. Changed to the suggested `value()`.

opentasks/src/main/java/org/dmfs/tasks/notification/signals/SwitchableSignal.java
```diff
@@ -0,0 +1,43 @@
...
+ * {@link NotificationSignal} that delegates to either {@link NoSignal} or {@link SettingsSignal} based on the received boolean.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SwitchableSignal implements NotificationSignal
+{
+    private final NotificationSignal mDelegate;
+
+
+    public SwitchableSignal(Context context, boolean withoutSignal)
```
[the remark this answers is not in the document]
lemonboston: Updated.

opentasks/src/main/res/values/strings.xml
lemonboston: Added them.

lemonboston:
Ready for next round of review.

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * Unit test for {@link Toggled}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class ToggledTest
+{
+    @Test
+    public void testValidFlags()
+    {
+        new Toggled(Notification.DEFAULT_SOUND, true, new NoSignal());
```
dmfs: This test doesn't test anything. It just creates an object but doesn't use it. Why would I do that in my code?

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testInValidFlag()
+    {
+        new Toggled(15, true, new NoSignal());
+    }
+
+
+    @Test
+    public void testAddingFlag()
+    {
+        assertTrue(containsFlag(
```
dmfs: don't use `containsFlag` here, instead test the expected outcome like so

```java
assertThat(new Toggled(Notification.DEFAULT_SOUND, true, new NoSignal()).value(), is(new NoSignal().value() | Notification.DEFAULT_SOUND));
assertThat(new Toggled(Notification.DEFAULT_SOUND, false, new NoSignal()).value(), is(new NoSignal().value()));
```

opentasks/src/main/java/org/dmfs/tasks/utils/BitFlagUtils.java
```diff
@@ -0,0 +1,56 @@
...
+package org.dmfs.tasks.utils;
+
+/**
+ * Convenience methods for manipulating bit flags.
+ * <p>
+ * Code from <a href="https://medium.com/@JakobUlbrich/flag-attributes-in-android-how-to-use-them-ac4ec8aee7d1">Flag Attributes in Android — How to Use Them</a>
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class BitFlagUtils
```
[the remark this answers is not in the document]
dmfs: Just noticed that not even `ToggledTest` should use any of these methods.

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
```diff
@@ -0,0 +1,62 @@
...
+        new Toggled(Notification.DEFAULT_SOUND, true, new NoSignal());
+        new Toggled(Notification.DEFAULT_VIBRATE, true, new NoSignal());
+        new Toggled(Notification.DEFAULT_LIGHTS, true, new NoSignal());
+    }
+
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testInValidFlag()
+    {
+        new Toggled(15, true, new NoSignal());
```
dmfs: Same here. If I never call `value()` why shouldn't I create a `Toggled` object with invalid params? Who cares if I don't use the value it returns?
A test case should test actual usage. But this test doesn't use the object it just creates one.

It's like building a car with just three wheels. That's totally fine if you don't intend to drive it, right?

This probably boils down to the question if a constructor should be code free (including validation) or not and whether eager validation is better than lazy validation. I think the latter one is more "declarative". I mean I can declare a trunk to be a car and that's fine if I don't intend to start the engine and actually drive with it (that's why it works pretty well for kids). If I try to start the engine of a trunk I'll fail badly, but not before.
In our code we don't care if the car is actually a trunk as long as we can start the engine. That's why we have test cases.
Or to go with the three wheel car example, we don't care how many wheels it has if it passes the "drive" test. Or do we?

opentasks/src/main/java/org/dmfs/tasks/notification/signals/SwitchableSignal.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SwitchableSignal implements NotificationSignal
+{
+    private final Context mContext;
+    private final boolean mWithoutSignal;
+
+
+    public SwitchableSignal(Context context, boolean withoutSignal)
```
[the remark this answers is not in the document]
lemonboston: You're right, it's better that way, I missed the decoration the first time. Updated now.

opentasks/src/main/java/org/dmfs/tasks/utils/BitFlagUtils.java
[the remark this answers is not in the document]
lemonboston: Removed the class and just used private methods as suggested.

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
lemonboston: Updated.

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
lemonboston: Moved this validation from ctor to method. Not sure we even need it, but it doesn't hurt. Updated the tests accordingly to call on `value()`.
(Personally, I don't have a clear opinion on the eager or lazy validation, in general I think fail-fast is good, but this declarative oop approach doesn't follow that, so not sure what fits best.)

opentasks/src/test/java/org/dmfs/tasks/notification/signals/ToggledTest.java
lemonboston: Updated.

lemonboston:
Read for review again.

Issue #407, opened 2017-08-18
GET_ACCOUNTS runtime permission handling


lemonboston:
https://github.com/dmfs/opentasks/issues/404#issuecomment-323271866 :

> @lemonboston the app should check for the GET_ACCOUNTS permission on every start. If the permission has been granted it should continue, otherwise it should show a message about this issue and a "Grant" button which then launches the permission request. If the permission has been denied with "don't ask again" the message should instruct the user to go to settings and grant the permission.
> 
> Also check if we're using other runtime permissions.

lemonboston:
Besides GET_ACCOUNTS the only other dangerous permission I see in the manifest is READ_CONTACTS. They both belong to the CONTACT group so once the user grants that, both will be granted.
But documentation say we shouldn't depend on these groups, so is the READ_CONTACTS permission needed to be checked on every start as well or is there a more specific place for that?

---

Just to have a better picture, where are these permissions used in the code? I mean that there is usually a lint error when trying to use an Android API which requires dangerous permission but we don't have that error, so is it because `ContentProvider` and/or `AccountManager` are too general to detect this?

---

I suppose `BaseActivity` `onCreate()` can be an obvious place to put this check. That is only introduced in this not yet merged pull request: https://github.com/dmfs/opentasks/pull/388
I will start to work on top of that, but if that doesn't make it to master, I can rebase of course and place the code in current Activities.

dmfs:
We don't use READ_CONTACTS yet. We've already added this permission in advance for future features (like task delegation). So at present there is no need to request it.
The code which uses `GET_ACCOUNTS` is in OpenTasksProvider (and only there). It's quite possible that there is no lint error, because the function we use still works without this permission. It just has a different behavior (it only returns accounts which are owned by the app instead of all accounts).

dmfs:
Released in 1.1.10

PR #409, opened 2017-08-25
Requesting GET_ACCOUNTS permission on startup. #407

lemonboston:
@dmfs I am submitting this now so you can check and decide on UX steps and the exact messages, but the code is not finalized here. I will get back to this when the permission handling library is ready. (Or let me know if it's better to not wait for that.)

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/utils/BaseActivity.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/utils/BaseActivity.java
@@ -16,6 +16,7 @@
 
 package org.dmfs.tasks.utils;
 
+import android.Manifest;
 import android.content.SharedPreferences;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
@@ -23,6 +24,9 @@ import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
 
 import org.dmfs.android.retentionmagic.RetentionMagic;
+import org.dmfs.tasks.utils.permission.BasicAppPermissions;
+import org.dmfs.tasks.utils.permission.Permission;
+import org.dmfs.tasks.utils.permission.dialog.PermissionRequestDialogFragment;
 
 
 /**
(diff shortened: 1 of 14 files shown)


opentasks/build.gradle
```diff
@@ -49,5 +49,6 @@ dependencies {
     compile project(':opentasks-provider')
     compile 'com.google.android.apps.dashclock:dashclock-api:2.0.0'
     compile 'com.github.dmfs:color-picker:1.0'
+    compile 'com.github.dmfs.jems:iterators:1.7.1'
```
dmfs: Just use `org.dmfs:iterators:1.7.1`. There is no need to specify the github jems version. In fact, the maven dependency name will remain the same even with the jems version. Same goes for `optional`.

lemonboston: Ok, updated.

dmfs:
Ok. It's a start, but we need to tweak it here and there. Here are some initial notes:

* The description of the permission request is not clear or even wrong. It sounds as if OpenTasks would sync tasks which is not the case. Also there should be a link to the contacts permission issue.
    
    I'll work on a better text.
* If the UI finds any synced lists, it should not offer the option to not grant the permission. I.e. if we know that the user has synced lists, the permission is no longer optional. Just check if there are any task lists with an `account_type` other than the local type.

dmfs:
"don't show again" only means don't show again as long as there are no synced lists. If there is at least one synced list, the dialog should pop up again if it has been dismissed initially.
Also, the dialog should be modal. It should not be possible to dismiss it by taping on the background.

lemonboston:
I see, I have two questions.
OpenTaskPal would be useful for querying the task lists, it's not that important, but if this task is not urgent, can we wait until the tests task is complete so I could use OpenTaskPal here?
Would it be okay to start to use RxJava2 here for that query? I think it could be beneficial in general to start to use Rx for loadings with that kind of declarative async access approach we chatted about.

TODOs:

- [ ] Create and publish permissions library (Marten)
- [ ] Integrate permission library, possibly improve related design here (Gabor)
- [ ] Better dialog description text (Marten)
- [ ] Don't offer 'don't show again' button when there is any synced list (Gabor)
- [ ] Show dialog again when 'don't show again' had been selected, but synced list found (Gabor)
- [ ] Dialog should not be dismissible by tapping on background (Gabor)

dmfs:

lemonboston:
Rebased on master.

My idea is to leave this on hold until permission lib is available because that can be a main driver of the code structure here. But let me know if I should continue working on the parts which are not closely related nevertheless.

dmfs:
@lemonboston please rebase again.

lemonboston:
Rebased and checked the workings again. Apart from the messages it looks good to me as a first approximation.
If user denies with "don't show again" at the permission grant dialog, but not at ours, then after tapping OK on ours, nothing happens. But that's kind of an edge case.

Also updated the dialog to not be dismissable by tapping outside.

dmfs:
Please replace the dialog text with
```
In order to read the list of accounts on your device which support task synchronization, OpenTasks needs the GET_ACCOUNTS permission.

Please note that (due to the great wisdom of the Android developers) the permission request will be labeled as "access your contacts" instead of "read your accounts". See <a href="https://issuetracker.google.com/issues/37065963">Issue 37065963</a> for further information.
``'

lemonboston:
Updated the dialog, ready for check again.

lemonboston:
As requested, I've remove the opt out option for granting the permission. It's forced on the user now, checked in `BaseActivity.onResume()`. I've also added handling the 'don't ask again' case, for that the button changes to `OPEN SETTINGS` and opens the phone settings for the app when clicked.
That could have a slightly different message, but I guess that's not important.

dmfs:
@lemonboston please rebase.

lemonboston:
Rebased.

dmfs:
Merging this despite the CI issue. It works for me.

Issue #410, opened 2017-08-29
Create unit test for opentasks-provider


dmfs:
In order to publish support for recurring tasks we need to test the provider carefully.

Issue #392, opened 2017-07-26
Drop Android 2.x and 3.x support.


dmfs:
Market shares of Android 2.x and 3.x are close to zero. So it's hard to justify continued support and efforts for these Android versions.

This issue is about removing all workarounds we had to introduce for these versions as well as obsolete styles.

lemonboston:
When this is implemented, make sure to look through all AndroidManifests in the project for
```
<uses-sdk tools:overrideLibrary=...>
```
usage, since those have been added because the used library had higher `minSdkVersion`. So those can possibly be removed at that time.

dmfs:
@lemonboston The new `minSdkLevel` is `15`. You don't need to spend much time on removing all the workarounds for Android 2 and 3. Maybe just remove a few big annoying ones. We'll take care of the others whenever we stumble across them.

lemonboston:
@dmfs Could we abandon API 15 - Android 4.0 - Ice Cream Sandwich as well?
According to the official Dashboard, it has only 0.5% share:
https://developer.android.com/about/dashboards/index.html

Since 4.1 Jelly Bean was a major, annual version update after 4.0, I suppose there are framework APIs that changed that time, and we could potentially remove some related backward compatible code for that as well.
https://en.wikipedia.org/wiki/Android_version_history (4.0: October 18, 2011, 4.1: July 9, 2012)

dmfs:
We won't drop support for 15 atm. I can't remember any limitations or workarounds we had to add to support API level 15.
If you come across any, please open an issue.

lemonboston:
Ok.

PR #418, opened 2017-09-12
Tests for TaskProvider #410

lemonboston:
Remove dynamic authority resolution in AuthorityUtil.

---

Submitting this now for first check. There are many TODOs left in code.

A few specific notes:
- opentaskpal classes are not reviewed yet
- the contentpal test support ended up being one simple `Matcher` class basically, but I used this `Check` type to implement it and make an adapter for `Matcher`. I haven't thought it through yet since things got much more simple with the `Assert` operations. One thing is that `Matcher` seems not to be completely straighforward for me with the 3 base classes, also I think what we do here is more an execution which can throw Exception than a simple comparison of an already calculated value, so that's why this interface may be better.
We could also forget the hamcrest assert as well, I suppose, to write something like:
```
execute(batch).with(client).andVerifyContent(assertOperations)
```
with our own types.
Implementation may depend a little bit on the eager queue stuff as well.
- not sure where the convenience `Assert*` composed operations should go, to contentpal module or contentpal-testing module
- I've left `AA_TO_*` package names to indicate which module the sibling packages are intended to end up in
- There is one bigger test case left in original form, I will update that, too, so the related private methods can be removed as well
- All test cases are based on the original ones from the old code
- There are possible ways to improve the test code usage, there are some issues to be solved as well (TODOs in code)
- `Assert` for `RowSnapshot` is not available yet, so that will result in changes as well

This is a single commit now, I will continue to commit on top of it.

### DIFF
--- a/opentasks-provider/build.gradle
+++ b/opentasks-provider/build.gradle
@@ -21,11 +21,21 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_7
         targetCompatibility JavaVersion.VERSION_1_7
     }
+    packagingOptions {
+        exclude 'META-INF/NOTICE'
+        exclude 'META-INF/LICENSE'
+    }
 }
 
 dependencies {
     compile('org.dmfs:rfc5545-datetime:0.2.4')
     compile('org.dmfs:lib-recur:0.9.6')
     compile project(':opentasks-contract')
-    testCompile 'junit:junit:4.12'
+
+    androidTestCompile project(':opentaskspal')
+    androidTestCompile 'com.github.dmfs.contentpal:contenttestpal:' + CONTENTPAL_VERSION
+    androidTestCompile 'com.android.support:support-annotations:' + SUPPORT_LIBRARY_VERSION
+    androidTestCompile 'com.android.support.test:runner:0.5'
+    androidTestCompile 'com.android.support.test:rules:0.5'
+    androidTestCompile 'org.hamcrest:hamcrest-library:1.3'
 }
--- a/opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
+++ b/opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
@@ -1,0 +1,403 @@
+/*
+ * Copyright 2017 dmfs GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dmfs.provider.tasks;
+
+import android.content.ContentProviderClient;
+import android.content.ContentResolver;
+import android.content.ContentValues;
+import android.content.Context;
+import android.os.Build;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.dmfs.android.contentpal.OperationsQueue;
+import org.dmfs.android.contentpal.RowSnapshot;
+import org.dmfs.android.contentpal.Table;
+import org.dmfs.android.contentpal.batches.MultiBatch;
+import org.dmfs.android.contentpal.batches.SingletonBatch;
+import org.dmfs.android.contentpal.operations.Assert;
+import org.dmfs.android.contentpal.operations.BulkDelete;
+import org.dmfs.android.contentpal.operations.Delete;
+import org.dmfs.android.contentpal.operations.Put;
+import org.dmfs.android.contentpal.operations.Referring;
+import org.dmfs.android.contentpal.predicates.AllOf;
+import org.dmfs.android.contentpal.predicates.EqArg;
+import org.dmfs.android.contentpal.queues.BasicOperationsQueue;
+import org.dmfs.android.contentpal.rowdata.Composite;
+import org.dmfs.android.contentpal.rowdata.EmptyRowData;
+import org.dmfs.android.contentpal.rowsnapshots.VirtualRowSnapshot;
+import org.dmfs.android.contenttestpal.operations.AssertEmptyTable;
+import org.dmfs.android.contenttestpal.operations.AssertRelated;
+import org.dmfs.opentaskspal.tables.InstanceTable;
+import org.dmfs.opentaskspal.tables.LocalTaskListsTable;
+import org.dmfs.opentaskspal.tables.TaskListScoped;
+import org.dmfs.opentaskspal.tables.TaskListsTable;
+import org.dmfs.opentaskspal.tables.TasksTable;
+import org.dmfs.opentaskspal.tasklists.NameData;
+import org.dmfs.opentaskspal.tasks.OriginalInstanceSyncIdData;
+import org.dmfs.opentaskspal.tasks.SyncIdData;
+import org.dmfs.opentaskspal.tasks.TimeData;
+import org.dmfs.opentaskspal.tasks.TitleData;
+import org.dmfs.rfc5545.DateTime;
+import org.dmfs.rfc5545.Duration;
+import org.dmfs.tasks.contract.TaskContract.Instances;
+import org.dmfs.tasks.contract.TaskContract.TaskLists;
+import org.dmfs.tasks.contract.TaskContract.Tasks;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.dmfs.android.contenttestpal.ContentMatcher.resultsIn;
+import static org.junit.Assert.assertThat;
+
+
+/**
+ * Tests for {@link TaskProvider}.
+ *
+ * @author Yannic Ahrens
+ * @author Gabor Keszthelyi
+ */
+@RunWith(AndroidJUnit4.class)
+public class TaskProviderTest
+{
+    private ContentResolver mResolver;
+    private String mAuthority;
+    private Context mContext;
+    private ContentProviderClient mClient;
+
+
+    @Before
+    public void setUp() throws Exception
+    {
+        mContext = InstrumentationRegistry.getTargetContext();
+        mResolver = mContext.getContentResolver();
+        mAuthority = AuthorityUtil.taskAuthority(mContext);
+        mClient = mContext.getContentResolver().acquireContentProviderClient(mAuthority);
+
+        // Assert that tables are empty:
+        OperationsQueue queue = new BasicOperationsQueue(mClient);
+        queue.enqueue(new MultiBatch(
+                new AssertEmptyTable<>(new TasksTable(mAuthority)),
+                new AssertEmptyTable<>(new TaskListsTable(mAuthority)),
+                new AssertEmptyTable<>(new InstanceTable(mAuthority))));
+        queue.flush();
+    }
+
+
+    @After
+    public void tearDown() throws Exception
+    {
+        /*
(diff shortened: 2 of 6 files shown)


opentaskspal/src/main/java/org/dmfs/opentaskspal/operations/Task.java
```diff
@@ -0,0 +1,57 @@
...
+import org.dmfs.android.contentpal.Operation;
+import org.dmfs.android.contentpal.RowSnapshot;
+import org.dmfs.android.contentpal.SoftRowReference;
+import org.dmfs.android.contentpal.TransactionContext;
+import org.dmfs.android.contentpal.operations.Related;
+import org.dmfs.optional.Optional;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+public final class Task implements InsertOperation<TaskContract.Tasks>
```
dmfs: The name is not ideal. This is a decorator which puts new a task on a specific list. The name should express that. Also it's valid to change the task list a list belongs to so this operation is not limited to inserts.

opentaskspal/src/main/java/org/dmfs/android/contentpal/testing/Check.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.dmfs.android.contentpal.testing;
+
+/**
+ * // TODO review javadocs
+ * Represents a check/matcher/condition/verification that can be checked against a target object of type <code>T</code>.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface Check<T>
```
dmfs: Do we really need this interface at all?

opentaskspal/src/main/java/org/dmfs/android/contentpal/predicates/DelegatingPredicate.java
```diff
@@ -0,0 +1,52 @@
...
+
+import android.support.annotation.NonNull;
+
+import org.dmfs.android.contentpal.Predicate;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public abstract class DelegatingPredicate implements Predicate
```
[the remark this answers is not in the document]
lemonboston: Yep, all class in the `org.dmfs.android.contentpal` package here are intended to end up there, so I start to create issue for them now. Done for this one.

opentaskspal/src/main/java/org/dmfs/android/contentpal/predicates/IdEq.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.dmfs.android.contentpal.RowSnapshot;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class IdEq<PrimaryTable> extends DelegatingPredicate
+{
+    public IdEq(final String foreignIdColumnName, final RowSnapshot<PrimaryTable> primaryRow, final String primaryIdColumnName)
```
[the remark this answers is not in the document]
lemonboston: You're right, that name is much better and there is no need to specify the primary id column.
Submitted pull request in contentpal: https://github.com/dmfs/ContentPal/pull/47

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/ListIdEq.java
```diff
@@ -0,0 +1,42 @@
...
+import org.dmfs.android.contentpal.predicates.AllOf;
+import org.dmfs.android.contentpal.predicates.DelegatingPredicate;
+import org.dmfs.android.contentpal.predicates.IdEq;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class ListIdEq extends DelegatingPredicate
```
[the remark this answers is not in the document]
lemonboston: Ok, renamed it to `TaskOnList`.

opentaskspal/src/main/java/org/dmfs/android/contentpal/testing/Check.java
lemonboston: Since we basically only have a single and simple `Matcher` now I suppose we don't, so I've removed it. I still think that it may be nice to have a custom flowable 'api' to test operations, but it's not important, we could get back to it later as well.

opentasks-provider/build.gradle
```diff
@@ -17,11 +17,26 @@ android {
...
+
+    androidTestCompile project(':opentaskspal')
+    androidTestCompile 'com.android.support:support-annotations:' + SUPPORT_LIBRARY_VERSION
+    androidTestCompile 'com.android.support.test:runner:0.5'
+    androidTestCompile 'com.android.support.test:rules:0.5'
+    androidTestCompile 'org.hamcrest:hamcrest-library:1.3'
+
+    // https://github.com/amitshekhariitbhu/Android-Debug-Database
+    // Usage: "adb forward tcp:8080 tcp:8080" then open "http://localhost:8080"
+    androidTestCompile 'com.amitshekhar.android:debug-db:1.0.1'
```
dmfs: What's this for and does it have to be in master?

opentaskspal/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,9 @@
+<manifest xmlns:tools="http://schemas.android.com/tools"
+        package="org.dmfs.opentaskspal">
+
+    <!--TODO Remove after https://github.com/dmfs/opentasks/issues/392-->
+    <uses-sdk tools:overrideLibrary="org.dmfs.android.contentpal"/>
```
dmfs: please add a note to #392 about this.

opentaskspal/src/main/java/org/dmfs/android/contentpal/operations/AssertNoRow.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @author Gabor Keszthelyi
+ */
+// TODO This can probably use the Assert that receives RowSnapshot when that's available, instead of querying for RowData
+public final class AssertNoRow<T> extends DelegatingOperation<T>
+{
+    public AssertNoRow(Table<T> table, RowData<T> rowData)
+    {
+        super(new Counted<T>(0, new BulkAssert<T>(table, rowData)));
```
dmfs: Again, this is the same as `AssertEmptyTable`. It always fails if the given `Table` is not empty.

opentaskspal/src/main/java/org/dmfs/android/contentpal/operations/AssertRow.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class AssertRow<T> extends DelegatingOperation<T>
+{
+
+    public AssertRow(Table<T> table, RowSnapshot<T> rowSnapshot, RowData<T> rowData)
+    {
+        super(new Counted<T>(1,
+                new BulkAssert<T>(table,
```
dmfs: This should use `Assert` now.

dmfs:
As mentioned before I see a general problem with the "RelatedTo" type of Predicates. They won't work if the row has not been loaded from the database. Because the `VirtualRowSnapshot` doesn't contain the `_ID` it got after it had been committed. That's why this type of constraint needs to be added by an `Operation` decorator, because it has access to the `TransactionContext`.

dmfs:
The test project should use a different provider authority like `org.dmfs.opentasks.test` to allow testing even if OpenTasks is already installed on the test device.

opentasks-provider/build.gradle
lemonboston: This was very useful for me, it allows to check the content of the db from browser. Do you maybe use something else for that?
Anyway, I've removed it, no need for it to be committed.

opentaskspal/src/main/AndroidManifest.xml
lemonboston: added: https://github.com/dmfs/opentasks/issues/392#issuecomment-329198715

opentaskspal/src/main/java/org/dmfs/android/contentpal/operations/AssertNoRow.java
lemonboston: This has been removed already.

opentaskspal/src/main/java/org/dmfs/android/contentpal/operations/AssertRow.java
lemonboston: This was removed as well while you were reviewing.

lemonboston:
Test not uses different authority name (added a resource file under `androidTest` to override the string resource).

opentaskspal/src/main/java/org/dmfs/android/contentpal/operations/AssertNoRow.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @author Gabor Keszthelyi
+ */
+// TODO This can probably use the Assert that receives RowSnapshot when that's available, instead of querying for RowData
+public final class AssertNoRow<T> extends DelegatingOperation<T>
+{
+    public AssertNoRow(Table<T> table, RowData<T> rowData)
+    {
+        super(new Counted<T>(0, new Assert<T>(table, rowData)));
```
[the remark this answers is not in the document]
lemonboston: (Addressed already)

opentaskspal/src/main/java/org/dmfs/opentaskspal/operations/Task.java
lemonboston: This comes back to the question of whether to keep `InsertOperation` and `Operation`, because if we keep them, than we need to make 2 classes here for the same thing since it cannot be used in `Table.insertOpertation()` for return vallue if it's an `Operation`.
For now, I will just leave it as an `InsertOperation` and rename it.

lemonboston:
(Corrected typo in previous comment: "Test not uses.." -> "Test now uses..")

lemonboston:
OpenTaksPal classes are ready for review (i.e. classes in `opentaskpal` module's `org.dmfs.opentaskspal` package. The ones in `org.dmfs.android.contentpal` package are all supposed to be replaced with ones from ContentPal.)

lemonboston:
This is now rebased on top of https://github.com/dmfs/opentasks/pull/422, I've changed the title as well.
`PostOperationAssert*` classes are still used here, but could be replaced with `BulkAssert(..ReferredTo())` from contentpal PR if needed as mentioned on Slack.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -0,0 +1,404 @@
...
+     */
+    @Test
+    public void testInstanceDelete()
+    {
+        RowSnapshot<TaskLists> taskList = new VirtualRowSnapshot<>(new LocalTaskListsTable(mAuthority));
+        Table<Tasks> taskTable = new TaskListScoped(taskList, new TasksTable(mAuthority));
+        RowSnapshot<Tasks> task = new VirtualRowSnapshot<>(taskTable);
+        OperationsQueue queue = new BasicOperationsQueue(mClient);
+
+        assertThat(new MultiBatch(
```
dmfs: There is no need to assert this again if it already has been verified by a previous test. Just insert the rows.

lemonboston: Removed that unnecessary assert.

dmfs:

lemonboston:
Rebased on master and updated tests with latest contentpal classes, finalized the test code, added javadocs for all test methods.

To sum up, this pull request adds Instrumentation tests for TaskProvider, based on test cases from earlier times of the project. (ticket for further tests: https://github.com/dmfs/opentasks/issues/420)
It also removes the dynamic authority resolution.

There are 2 TODOs left in the code, let me know if either is worth having a ticket:
- Test Orchestration is a coming feature which will allow each test method to run in its own instrumentation, hence removing the need for some code in setUp() and tearDown()
- Authority is now simply cached under the same class, AuthorityUtil, design or name could be improved (results in changes in many files)

dmfs:
Yes, please create tickets.

lemonboston:
Created the tickets for the TODOs: https://github.com/dmfs/opentasks/issues/437, https://github.com/dmfs/opentasks/issues/438

Issue #75, opened 2014-05-27
Feature Request: Reorder checklist by drag/drop


maulwuff:
It would be a nice feature to reoder a checklist styled task by drag/drop

Ezelty:
I know this issue is quite old, but I would really like to be able to reorder my items! If it's too complicated to add a drag/drop system, could you add an option to sort the items alphabetically? Or an option to set some items as more important, and put them on the top of the checklist? I often put a star or an emoji at the beginning of some items to see which items are important and create sort of categories, so having my items sorted alphabetically would be very useful to me,as it would gather the items with a star or with the same emoji. :)

dmfs:
Drag and drop shouldn't be too difficult. We just had to make some internal changes to make this work. The latest release already contains these internal improvements, so now it's "just" the UI that needs some work.

dmfs:
We've just finished the implementation of this and are currently testing it. Send us an email if you want to test the pre-release of the app.

maulwuff:
Gern

PR #422, opened 2017-09-14
OpenTaskPal module (ContentPal extension for TaskProvider). #421

lemonboston:
Note that these are only the classes and constructors that were needed for the test cases, so this is not a complete ContentPal support for TaskProvider at the moment. I think we can extend it later, for example when test coverage is improved, since test and opentaskpal will be in same project so both can be extended in the same commits.

Also note some TODOs are pending on ContentPal PRs.


lemonboston:
Missed to create `NoTimeData` that clears start, due, duration, added it now.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * {@link RowData} for setting only the due datetime of the task, for setting start as well use {@link TimeData}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class DueData extends DelegatingRowData<TaskContract.Tasks>
+{
+    public DueData(@NonNull Long dueTimeStamp)
+    {
+        super(new SimpleRowData<TaskContract.Tasks>(TaskContract.Tasks.DUE, dueTimeStamp));
```
dmfs: This should also set START and DURATION to null.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/NoTimeData.java
```diff
@@ -0,0 +1,43 @@
...
+{
+    @NonNull
+    @Override
+    public ContentProviderOperation.Builder updatedBuilder(@NonNull TransactionContext transactionContext, @NonNull ContentProviderOperation.Builder builder)
+    {
+        return builder
+                .withValue(TaskContract.Tasks.DTSTART, null)
+                .withValue(TaskContract.Tasks.DUE, null)
+                .withValue(TaskContract.Tasks.DURATION, null);
+    }
```
dmfs: A task without times is currently not allowed to recur, so RDATE, RRULE and EXDATE should also be set to null.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * {@link RowData} for setting only the due datetime of the task, for setting start as well use {@link TimeData}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class DueData extends DelegatingRowData<TaskContract.Tasks>
+{
+    public DueData(@NonNull Long dueTimeStamp)
```
dmfs: Date-time values should be typed `DateTime`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeZoneData.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.TimeZone;
+
+
+/**
+ * {@link RowData} for the time zone value of the task.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class TimeZoneData implements RowData<TaskContract.Tasks>
```
dmfs: Setting time zone alone doesn't make sense. Since it's part of `DateTime` it should be set by the other date & time `RowData` implementations.

dmfs: Ideally OpenTasksPal provides high-level access to the content provider. Check out `SingleEventData` in CalendarPal it takes `DateTime`s and automatically sets timestamps, time zones and all-day flags. Loading or setting a time zone or an all-day flag without loading or setting the timestamp too doesn't make much sense and is considered error prone.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
lemonboston: I understand the need to set DURATION to null, but wouldn't this class be used also to just update DUE, keeping START? Or for that with need to use `TimeData` and retrieve START first?

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeZoneData.java
lemonboston: There is only one timezone and all-day flag for tasks, I assume `start` and `due` with different timezone should be rejected, right? What about the all-day flag? Take it only from start, or validate if its the same for `start` and `due`?
Does it make sense to set all-day flag in `DueTime` from due?

dmfs: No different time zones should not be rejected. Actually we should update the provider to support this case. For now just shift start to the due time zone.
But start and due must have the same all-day flag.

lemonboston:
Updated the time `RowData`s, please check if they make sense. I would write unit test afterwards.
Also, there is a possibility to use that `Equalator` interface for the time zone comparison, is it okay to add it here now? (It could go to `jems` later.)

lemonboston:
Updated to latest contentpal master and removed all related files from here.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public ContentProviderOperation.Builder updatedBuilder(@NonNull TransactionContext transactionContext, @NonNull ContentProviderOperation.Builder builder)
+    {
+        if (mDue != null && mStart.isAllDay() != mDue.isAllDay())
+        {
+            throw new IllegalArgumentException("'start' and 'due' must have the same all-day flag");
+        }
+
+        DateTime start = mStart;
+        if (mDue != null && !TimeZoneEqualator.INSTANCE.areEqual(mStart, mDue))
```
dmfs: There is no need for this Equalator. DateTime is smart enough to return the same instance if the timezone is considered to be the same.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
```diff
@@ -0,0 +1,59 @@
...
+        mDue = due;
+    }
+
+
+    @NonNull
+    @Override
+    public ContentProviderOperation.Builder updatedBuilder(@NonNull TransactionContext transactionContext, @NonNull ContentProviderOperation.Builder builder)
+    {
+        return builder
+                .withValue(TaskContract.Tasks.DUE, mDue.getTimestamp())
```
dmfs: Also set allday flag and time zone.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
```diff
@@ -0,0 +1,59 @@
...
+                .withValue(TaskContract.Tasks.DTSTART, null)
+                .withValue(TaskContract.Tasks.DURATION, null)
+
+                // A task without times is currently not allowed to recur:
```
dmfs: but this one has a time!? The point is that we can't update due without updating recurrence because updating due usually also affects recurrence.
For now it's ok to null the recurrence columns because recurrence is not really supported yet. But the reason stated in the comment doesn't apply here.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/NoTimeData.java
dmfs: Also wipe out all-day and time zone.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/OriginalInstanceSyncIdData.java
```diff
@@ -0,0 +1,40 @@
...
+
+
+/**
+ * {@link RowData} for adding {@link TaskContract.Tasks#ORIGINAL_INSTANCE_SYNC_ID}
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class OriginalInstanceSyncIdData extends DelegatingRowData<TaskContract.Tasks>
+{
+    // TODO Should this be @Nullable?
```
dmfs: No because you should not be allowed to clear this once it has been created.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SyncIdData.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * {@link RowData} for {@link TaskContract.Tasks#_SYNC_ID}
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SyncIdData extends DelegatingRowData<TaskContract.Tasks>
+{
+    // TODO Should this be @Nullable?
```
dmfs: No.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
```diff
@@ -0,0 +1,110 @@
...
+        return builder
+                .withValue(TaskContract.Tasks.DTSTART, start.getTimestamp())
+                .withValue(TaskContract.Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID())
+                .withValue(TaskContract.Tasks.IS_ALLDAY, start.isAllDay() ? 1 : 0)
+
+                .withValue(TaskContract.Tasks.DUE, due == null ? null : due.getTimestamp())
+
+                .withValue(TaskContract.Tasks.DURATION, duration)
+
+                // A task without times is currently not allowed to recur:
```
dmfs: this task has times.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TitleData.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * {@link RowData} for the title of the task.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class TitleData extends DelegatingRowData<TaskContract.Tasks>
+{
+    public TitleData(@Nullable CharSequence title)
```
dmfs: The title of a task should not be allowed to be `null`. You may set it to an empty string, but null doesn't make sense.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
lemonboston: You mean that there is no need to compare the `getId()` `String` value but use reference equality on the `getTimeZone()` `TimeZone`s? Like `if (timeZone1 == timeZone2)`?
Don't we depend on some undocumented internals of `DateTime` this way?
I've updated the code, removed the `Equalator` and added a comment about this behavior, but I think it would be clearer to check the `getId()` values. `TimeZone` doesn't provide `equals()`, so we are really depending on reference equality here which doesn't seem to be safe.
I can do that in private method if you are not consent with that `Equalator`.
(Although it could be easier to test that code in isolation and ideally just mock it out in (upcoming) `TimeZoneTest` if it could be injected there, but it couldn't.. so it may not be important.)

dmfs: No that's not what I meant. I've added a new comment to the new code. The comment above was a bit misleading.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
lemonboston: Yeah, that was copied from Slack, but it probably meant 'recurrence times' originally, should have added that... anyway, I've removed the comment.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/OriginalInstanceSyncIdData.java
lemonboston: Thanks, removed the TODO.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SyncIdData.java
lemonboston: Thanks, removed the TODO.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
lemonboston: Removed the comment.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TitleData.java
lemonboston: Ok, changed to `@NonNull`.

lemonboston:
I see, thanks, I've updated the code. Although, I think we still depend a bit on the knowledge of `DateTime` internals, for example that as I see it ignores `null` input, which is good in this case.
But the upcoming unit test of `TimeData` will check these variations, so I guess it's okay.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
lemonboston: Done.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/NoTimeData.java
lemonboston: Updated.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
lemonboston: (addressed)

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DueData.java
lemonboston: (addressed)

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeZoneData.java
lemonboston: (addressed)

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
lemonboston: (addressed)

lemonboston:
Waiting for the `ContentProviderOperation.Builder/ RowData` `Matcher` support now from `contentpal` to write the unit test(s), otherwise ready for review again (all comments addressed).

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/TaskOnList.java
```diff
@@ -0,0 +1,50 @@
...
+        super(new AllOf(predicate,
+
+                /*
+                TODO Potential issue here with the predicate for update, delete and assert operations
+                See https://github.com/dmfs/ContentPal/pull/47#issuecomment-329470827
+                One of these 2 pull requests may allow to solve this problem:
+                https://github.com/dmfs/ContentPal/pull/51
+                https://github.com/dmfs/ContentPal/pull/52
+                */
+                new EqArg(TaskContract.Tasks.LIST_ID, taskListRow.values().charData(BaseColumns._ID).value("-1"))));
```
dmfs: We'll wait for https://github.com/dmfs/ContentPal/pull/75 to implement this. The current approach is not reliable enough.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tables/TaskListScoped.java
```diff
@@ -0,0 +1,99 @@
...
+import org.dmfs.opentaskspal.operations.TaskListTask;
+import org.dmfs.opentaskspal.predicates.TaskOnList;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * A view onto the {@link TaskContract.Tasks} table which contains only tasks from a specific task list.
+ * Tasks created with {@link #insertOperation(UriParams)} will automatically be added to this task list.
+ * <p>
+ * Note, if you create a {@link View} (using {@link #view(ContentProviderClient, String...)}) with a virtual {@link RowSnapshot}, the {@link View} will always
```
dmfs: FYI: I've added https://github.com/dmfs/ContentPal/issues/81 to address this.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasklists/NameData.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        mName = name;
+    }
+
+
+    @NonNull
+    @Override
+    public ContentProviderOperation.Builder updatedBuilder(@NonNull TransactionContext transactionContext, @NonNull ContentProviderOperation.Builder builder)
+    {
+        return builder.withValue(TaskContract.TaskLists.LIST_NAME, mName.toString());
```
dmfs: Why not delegate to a `CharSequenceRowData` here?

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/OriginalInstanceSyncIdData.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * {@link RowData} for adding {@link TaskContract.Tasks#ORIGINAL_INSTANCE_SYNC_ID}
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class OriginalInstanceSyncIdData extends DelegatingRowData<TaskContract.Tasks>
+{
+    public OriginalInstanceSyncIdData(@NonNull String originalInstanceSyncId)
```
dmfs: OriginalInstanceSyncId could also be just a `CharSequence`. So you probably should delegate to `CharSequenceRowData` instead.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SyncIdData.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * {@link RowData} for {@link TaskContract.Tasks#_SYNC_ID}
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SyncIdData extends DelegatingRowData<TaskContract.Tasks>
+{
+    public SyncIdData(@NonNull String syncId)
+    {
+        super(new RawRowData<TaskContract.Tasks>(TaskContract.Tasks._SYNC_ID, syncId));
```
dmfs: Same here, could be a `CharSequence`

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+
+    public TimeData(@NonNull DateTime start, @NonNull DateTime due)
+    {
+        this(start, due, null);
+    }
+
+
+    public TimeData(@NonNull DateTime start, @NonNull String duration)
```
dmfs: `Duration` should be passed as a `Duration` object not as a `String`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @author Gabor Keszthelyi
+ */
+// TODO Unit test
+public final class TimeData implements RowData<TaskContract.Tasks>
+{
+    @NonNull
+    private final DateTime mStart;
+    @Nullable
+    private final DateTime mDue;
```
dmfs: Maybe make them (`mDue` and `mDuration`) `Optional` instead.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasklists/NameData.java
lemonboston: `CharSequenceRowData` allows `null`, so `NameData` wouldn't fail with that. Maybe we need another class for that. ?

lemonboston: Hmm.. I am not sure of the correct approach here actually. We don't validate for `null` just use the `@Nullable` `@NonNull` annotations. Should that be enough in these cases? I mean if there wasn't a `.toString()` call here, it wouldn't fail either without a validation.
Similar question come up with `OriginalInstanceSyncIdData` below.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
lemonboston: Right, I've changed it. We still use `@Nullable` at the ctor params? That's what I am not sure of, I don't see the end usage - is it going to be part of an 'automatic' syncronization mapping where it is good if 'null' can be passed on, or is the 'normal' usage, like how I would use the class from UI code is what we should aim for? In this latter case, I would put `Optional` in the constructor as well.

dmfs: No, we should not use null at all here. 

it should look like this:
```java
  private TimeData(@NonNull DateTime start, @NonNull Optional<DateTime> due, @NonNull Optional<Duration> duration)
    {
        mStart = start;
        mDue = due;
        mDuration = duration;
    }


    public TimeData(@NonNull DateTime start, @NonNull DateTime due)
    {
        this(start, new Present<>(due), absent());
    }

…
```

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasklists/NameData.java
dmfs: apparently we need another delegate which doesn't allow `null`. Or make sure `CharSequenceRowData` doesn't allow `null` and find another, better solution to "erase" values in the database.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/TimeData.java
lemonboston: Ok, updated.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasklists/NameData.java
lemonboston: This comes back to the topic of whether to delete or leave a value alone when the incoming update doesn't have it (which may be represented as `null`). Because that's what makes using `Optional` instead of the `null`s ambiguous as you mentioned before.
I don't know the actual usage in syncadapters..   can the operations maybe be distinguished based on whether they are in 'override' mode or 'fill in the missing data' mode?
If so, could this maybe be a flag? So we could either have an extended `Optional` model with that flag or pass it in the constructor as separate parameter.

lemonboston: Updated the class with a TODO about this nullability.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/OriginalInstanceSyncIdData.java
lemonboston: Updated the class with a TODO about this nullability.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SyncIdData.java
lemonboston: Updated the class with a TODO about this nullability.

lemonboston:
@dmfs Could you please help a little with this?:
Can there be a nullpointer here?
```
withValue(TaskContract.Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID())
```
I am not sure of the relationship between all day and the timezones and the ways `DateTime` can be constructed...   can `getTimeZone()` return `null` here?
Thanks.

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/TaskOnList.java
dmfs: #75 has been merged.

lemonboston:
About the DateTime timezone<->all-day flag question, I figured they are mutually exclusive, right?

I've update to latest contentpal and added unit tests. Let me know if I should squash.
The only outstanding TODOs are about the nullability, otherwise ready for review.

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/TaskOnList.java
lemonboston: Yes, I've replaced this with the `ReferringTo(Predicate)`.

dmfs:
Technically you could have a non-allday task without time zone. That's called floating time. But OpenTasks doesn't really support that so it's ok for now. Maybe create an issue to remind us about this.

Same goes for the nullable. We'll keep it like this and revisit later on when we may have a better idea of how to handle it.

lemonboston:
Ok, I've created the tickets:
https://github.com/dmfs/opentasks/issues/431
https://github.com/dmfs/opentasks/issues/432

Issue #279, opened 2016-10-03
Settings


dmfs:
Many of the current feature requests require a proper settings screen.

lemonboston:
An earlier note about this story from @dmfs:

> Regular PreferenceFragment, on Android 2.x it's not supported, just disable it.
> Add to settings:
> - show notification or not
> - notification sound (maybe allow to set independently for start and due)
> - vibrate on notification or not

lemonboston:
@dmfs I've submitted a PR https://github.com/dmfs/opentasks/pull/389 with a Settings screen containing only "show notification or not".
Is continuing with this a priority now and if so what should be the next settings to add?

dmfs:
We've received a couple of requests to change the notification sound (#124) and disable notification vibration (#305).
These should be next.

PR #424, opened 2017-09-14
merge Settings/master

dmfs:
@lemonboston please resolve conflicts


lemonboston:
I've squashed and rebased on master but I haven't pushed since I am not sure this is what you wanted.. do you want to keep the 2 commits is okay to force push the squashed, rebased, resolved version with message: "Settings screen with Notification singal settings. #279 #305" ?

dmfs:
We should keep these two commits separate. Only the commit which resolves the conflict should be squashed with the commit which caused the conflict.

lemonboston:
Ok, I've rebased, resolving conflicts without squashing.

Issue #440, opened 2017-10-06
Build of opentaskspal fails


dmfs:
Apparently `packagingOptions` need to be added to the gradle build file:

```
Execution failed for task ':opentaskspal:transformResourcesWithMergeJavaResForDebugAndroidTest'.
> com.android.build.api.transform.TransformException: com.android.builder.packaging.DuplicateFileException: Duplicate files copied in APK META-INF/LICENSE
        File1: /home/marten/.gradle/caches/modules-2/files-2.1/org.dmfs/rfc5545-datetime/0.2.4/204c1caf63c43c54d803282aa7a15bb0b0e5304e/rfc5545-datetime-0.2.4.jar
        File2: /home/marten/.gradle/caches/modules-2/files-2.1/org.dmfs/optional/1.8/c2bec94eaafb1e08a514feb6466de9f6086a3421/optional-1.8.jar
        File3: /home/marten/.gradle/caches/modules-2/files-2.1/org.dmfs/iterators/1.8/2065553fb0b4d67963318260784b331c28aaf41a/iterators-1.8.jar
```

PR #441, opened 2017-10-09
Exclude META-INF in opentaskpal gradle packagingOptions to fix CI bui…

lemonboston:
…ld. #440

---

I've added the packagingOptions exclude, although it had built for me without it as well, so I don't know if this fixes the issue. Was it a maven build task or Travis that failed?


dmfs:
IIRC it was `gradle check`

Issue #450, opened 2017-10-13
OpenTaskPal support for subtasks related operations and queries


lemonboston:

dmfs:
Released in 1.1.10

PR #452, opened 2017-10-13
Update contentpal version to latest. #451

lemonboston:

PR #453, opened 2017-10-13
OpenTaskPal subtask operations and queries support. #450

lemonboston:
I've added initial classes for the subtask relation support. @dmfs Could you please have a first look?
Please suggest better names, constructor parameter combinations, new classes or type of operations, etc, that you think we will need.
There are 3 classes in a temporary `contentpal` package, those are intended to be created in ContentPal.


opentaskspal/src/main/java/org/dmfs/opentaskspal/contentpal/tables/Filtered.java
```diff
@@ -0,0 +1,85 @@
...
+import org.dmfs.android.contentpal.Table;
+import org.dmfs.android.contentpal.UriParams;
+import org.dmfs.android.contentpal.View;
+import org.dmfs.android.contentpal.predicates.AllOf;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class Filtered<T> implements Table<T>
```
dmfs: I'm not sure if a `Filtered` `Table` decorator is a good idea, because it changes the semantics. With the "scoped" tables we already have you can insert rows and you always update and delete the same rows and every `View` it returns will also return the rows you inserted.
That's not true for `Filtered` anymore. I can insert a row, but I can't update it if it doesn't pass the filter, nor can I read it from the `View` this table returns in that case.

opentaskspal/src/main/java/org/dmfs/opentaskspal/operations/SubTasking.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class SubTasking extends DelegatingOperationsBatch
+{
+
+    public SubTasking(String authority, RowSnapshot<TaskContract.Tasks> subtask, RowSnapshot<TaskContract.Tasks> parentTask)
+    {
+        super(new MultiBatch(
+                new Put<>(subtask, new Parent(parentTask)),
```
dmfs: It's not necessary to update `PARENT_ID` and the relations. `PARENT_ID` is updated automatically.

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/Relation.java
```diff
@@ -0,0 +1,32 @@
...
+package org.dmfs.opentaskspal.predicates;
+
+import org.dmfs.android.contentpal.predicates.DelegatingPredicate;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class Relation extends DelegatingPredicate
```
dmfs: The name doesn't really express that this is a predicate. Consider calling it `IsRelation` or something like this. Same goes for `Property`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tables/RelationsTable.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * {@link Table} for {@link TaskContract.Property.Relation}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class RelationsTable extends DelegatingTable<TaskContract.Properties>
+{
+    public RelationsTable(String authority)
+    {
+        super(new Filtered<>(new Relation(), new PropertiesTable(authority)));
```
dmfs: This related to the issue with `Filtered` I mentioned earlier. If I insert something into a `RelationsTable` I'd expect that it automatically makes it a relation (or that it throws if I try to insert something that's not a relation).
Neither of these is the case here. Actually it looks like you never set the row mimetype when you insert into this table. That's fine if the table sets it (which is what I would expect) but a `Filtered` table doesn't set it either.

opentaskspal/src/main/java/org/dmfs/opentaskspal/views/TasksView.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class TasksView extends DelegatingView<TaskContract.Tasks>
+{
+    public TasksView(String authority, ContentProviderClient client)
+    {
+        super(new BaseView<TaskContract.Tasks>(client, TaskContract.Properties.getContentUri(authority)));
```
dmfs: This creates a `Properties` table. Copy'n'paste? ;-)

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SubTaskRelation.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @author Gabor Keszthelyi
+ */
+public final class SubTaskRelation extends DelegatingRowData<TaskContract.Properties>
+{
+    public SubTaskRelation(RowSnapshot<TaskContract.Tasks> subTask, RowSnapshot<TaskContract.Tasks> parentTask)
+    {
+        super(new Composite<>(
+                new Relation(subTask, TaskContract.Property.Relation.RelType.PARENT, parentTask),
```
dmfs: it's not necessary to set this on both tasks. In general it's preferred to set the parent relation on the subtask only. But you may have to make sure that any other parent relation with another task is removed first. Although we probably could do that automatically in provider code.

opentaskspal/src/main/java/org/dmfs/opentaskspal/contentpal/tables/Filtered.java
lemonboston: Ok, I've removed `Filtered` `Table` and `View`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/operations/SubTasking.java
lemonboston: Ok, removed the setting of `PARENT_ID`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/Relation.java
lemonboston: Update to `IsRelation` and `IsProperty`, but I also see that we don't need these currently. Shall we keep them or remove?

opentaskspal/src/main/java/org/dmfs/opentaskspal/tables/RelationsTable.java
lemonboston: As mentioned above, I've removed `Filtered` `Table` and `View`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/views/TasksView.java
lemonboston: Yep, good catch, thanks, fixed it.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SubTaskRelation.java
lemonboston: Ok, I've modified it to only set one direction of the relation.

What to do with removing other parent relations? Create ticket for implementing in provider then?

dmfs: Yep. Would be nice if we can offload some of the business logic into the provider and make the UI part easier.

lemonboston:
I've made changes, ready for check again. What else do we need atm?
(I've squashed multiple commits and force pushed so you may need to reset the branch)

opentaskspal/src/main/java/org/dmfs/opentaskspal/predicates/Relation.java
dmfs: Keep them.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/RelationData.java
```diff
@@ -0,0 +1,43 @@
...
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * {@link RowData} for adding a {@link TaskContract.Property.Relation} property.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class RelationData extends DelegatingRowData<TaskContract.Properties>
+{
```
dmfs: For some use cases (i.e. syncing) and also to test the provider we also need a ctor which takes a task UID (in form of a `CharSequence`) instead of a `RowSnapshot`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SubtaskRelationData.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+/**
+ * {@link RowData} to add the 'subtask relation' data about the 2 given tasks, to the {@link TaskContract.Properties} table.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class SubtaskRelationData extends DelegatingRowData<TaskContract.Properties>
+{
+    public SubtaskRelationData(RowSnapshot<TaskContract.Tasks> subTask, RowSnapshot<TaskContract.Tasks> parentTask)
```
dmfs: same here, sometimes you only have the `UID` of the parent or child task

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/RelationData.java
lemonboston: Just to confirm before I do anything, does it mean that we need to look up the `RowSnapshot` with something like this?:
```
new QueryRowSet(view, new HasUid(uid)).iterator().next()
```
Or, since sync uid is a unique id as well, would maybe the provider support looking the row up with url? Like with the standard id.

dmfs: You don't need to look it up. It's part of the contract that you can either pass an id, a uid or a uri : https://github.com/dmfs/opentasks/blob/master/opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java#L1494
The provider will look up the task and insert the id if none is given (see https://github.com/dmfs/opentasks/blob/master/opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java#L115)
Note that the UID can belong to a task which has not been synced yet or even to another calendar object (like and event or journal entry).

lemonboston: Added the extra constructor.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/SubtaskRelationData.java
lemonboston: I've replaced this class with two: `ParentTaskRelationData` and `ChildTaskRelationData`.
I guess it depends on the decision of how to represent the relations in the db whether these can be (both) useful in these forms.

lemonboston:
Read for check again.

dmfs:
Please rebase this on master.

dmfs:
Oops already merged it. I guess you have to resubmit #452

lemonboston:
Submitted as https://github.com/dmfs/opentasks/pull/465

PR #466, opened 2017-10-18
Add .travis.yml. Implements #417

dmfs:
This also changes the version code to be based on commit count in the current branch rather than master. Apparently Travis doesn't fetch the master which caused this to fail.
Since we only publish master builds this should not have any unwanted side effects. Yet we spread the version code to allow inserting version numbers if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: android
+sudo: false
+
+android:
+  components:
+    - tools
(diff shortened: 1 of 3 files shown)


dmfs:
Yeah, I'm not sure why the push build failed. Maybe there is some sort of timeout issue. We'll see how it works.

Issue #474, opened 2017-11-13
SmartView for showing progress as horizontal overlay on the background


lemonboston:
The code is duplicated over the `By*` classes now, extract and re-use.

PR #475, opened 2017-11-13
Progress background SmartView. #474

lemonboston:


lemonboston:
Build fails with
```
com.android.build.api.transform.TransformException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.dex.DexException: Multiple dex files define Lorg/dmfs/iterables/ArrayIterable
```

Iterators dependency insight:
```
org.dmfs:iterators:1.7.1 (conflict resolution)
\--- com.github.dmfs.androidcarrot:androidcarrot:5a4baa0222
     \--- compile

org.dmfs:iterators:1.5 -> 1.7.1
\--- au.com.codeka:carrot:2.4.0
     \--- compile

```

Plus `jems:1.13` is newly imported here.

lemonboston:
So I suppose we need to:
- exclude `iterators` from both codeka.carrot and dmfs.carrot
- build, test if sharing works runtime as well
- create tickets-PRs to update codeka.carrot and dmfs.carrot to latest jems (shall we instead replace iterators usage with JDK classes in codeka.carrot?)

Is this right?

dmfs:
That's right. Not sure if we can switch to jems in carrot. He wasn't quite happy about importing any library at all. I think this would only be an option if we can show big improvements for the existing code as well. Otherwise it might be easier to just copy the classes we use or replace the implementation with something less "sophisticated".

dmfs:
Implementation-wise this PR looks good. Have you tried letting `TaskFieldAdapters.PERCENT_COMPLETE` return an `Optional` instead of wrapping the result every time?

lemonboston:
I've checked that now and `TaskFieldAdapters.PERCENT_COMPLETE` is an `IntegerFieldAdapter`, so we would either need an `OptionalIntegerFieldAdapter` or general `Optional(FieldAdapter)` but as I see it's not trivial how to do it with current `FieldAdapter`:
```
public abstract class FieldAdapter<Type>
{
  public abstract Type get(ContentSet values);

  public abstract void set(ContentValues values, Type value);

  ....
}
```
in `get()` we could easily wrap in `NullSafe<T>` but in `set()` would need the actual value. In other words would the `FieldAdapter` have `Optional<T>` or `<T>` as type parameter? Having `Optional<T>` would mean adding `new Present<>()` for every `set()` usage.
So if this is something worth attending to, I think we should have a dedicated ticket.

I've excluded `iterators` from the carrot dependencies, the build succeeds now. I've also tested that sharing still works runtime. So this is ready for review.

dmfs:
Ok. Leave it then. We'll probably replace it with ContentPal sooner or later so it's not worth spending too much time with it now.
I'll check out the code.

dmfs:
Please update AndroidCarrot dependency. Looks good otherwise.

lemonboston:
I planned to rebase this on master after https://github.com/dmfs/opentasks/pull/478 is merged.

dmfs:
@lemonboston please rebase.

lemonboston:
Squash-rebased.

dmfs:
Despite the CI issue it works for me. So I'm going to merge this.

Issue #476, opened 2017-11-13
DateTime support in DateFormatter


lemonboston:
In order to aid and prepare for replacing `android.text.Time` with `org.dmfs.rfc5545.DateTime` in the project, as a developer, I would like the `DateFormatter` class to support `DateTime`s.
---

Implementation will need to convert between the 2 types.

dmfs:
Done

PR #479, opened 2017-11-16
DateTime support in DateFormatter. #476

lemonboston:
@dmfs Could you please review the conversion and Time-DateTime equivalence check in the test extra thoroughly?
You know the domain better, to assess that this conversion is good for the project or not.


dmfs:
your tests don't seem to validate the actual `Time` values, i.e. `minute`, `day', etc. Also I think your all-day handling might be incorrect. When an event is allday the timezone should be set to `UTC`. I'm not exactly sure what `DateTime` returns in that case, maybe it already returns `UTC`.

lemonboston:
I've added checking the individual fields in the test.

Also updated conversion to use "UTC" for all-day. But please note/check that `Time` constructors are used in different ways in the project, so if I searched for them, both the paremterless and with the timezone, so this "UTC" as default didn't look totally consistent from the first look. But I haven't checked the details.

Also rebased on master.

Ready for check again.

dmfs:
Merging this despite the CI issue.

Issue #472, opened 2017-10-31
Update project for Android Studio 3.0 and gradle plugin 3.0.0


lemonboston:
Android Studio 3.0 release announcement:
https://android-developers.googleblog.com/2017/10/android-studio-30.html
Migration guide for the gradle changes:
https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html

dmfs:
done

PR #481, opened 2017-11-17
Update gradle and buildTools to latest. Suppress a new Lint error. #472

lemonboston:
This updates gradle to latest. I've also added `google()` repository and a 'clean' task in the root gradle file which is added for some reason in new projects.

Lint marked this as error afterwards:
image

I suppressed it because I don't know what is expected there. @dmfs could you please check that? Can that binary flag operation be simplified or expressed differently?


dmfs:
The update caused a problem with CI: 
```
Checking the license for package Android SDK Build-Tools 26.0.2 in /usr/local/android-sdk/licenses

Warning: License for package Android SDK Build-Tools 26.0.2 not accepted.
```
Please make sure to update .travis.yml as well.

dmfs:
Regarding the Lint warning: Apparently Lint can't handle bit arithmetic. Otherwise it would know that this operations "unsets" the given flag.
You can ignore the warning. We'll probably replace this code eventually with something cleaner.

lemonboston:
I've updated the build tools version in `travis.yml` so the `push` build now pass but the `pr` one fails with:
```
:opentasks:connectedDebugAndroidTest
Skipping device 'test(AVD)' for 'opentasks:': Unknown API Level
 > : No compatible devices connected.[TestRunner] FAILED 
Found 1 connected device(s), 0 of which were compatible.
:opentasks:connectedDebugAndroidTest FAILED
```
Do you happen to know what this is? I see some configuration for an emulator in the yml, the version looks correct.

dmfs:
Not sure. Have you tried adding?
```
   - android-26
```

lemonboston:
One of the build still fails but due to https://github.com/dmfs/opentasks/issues/486
Otherwise this is ready for checking again.

PR #487, opened 2017-11-21
OpenTasksPal Task data reading classes for subtasks. #484

lemonboston:


lemonboston:
Added that `TaskUri`, ready for check.

lemonboston:
I've added constant fields for these data classes to expose the columns they require, like:
```java
public final class TaskColor implements Color
{
    public static final String[] PROJECTION = { TaskContract.Tasks.TASK_COLOR, TaskContract.Tasks.LIST_COLOR };
```

Since we have to know the projection in advance of actually having the data available and using these classes, I don't know if a method would work for this, because we don't have instances at that point.

I defined them as String arrays, maybe `Iterable` can work better, but then it needs to be converted to array before using it for `View.withProjection(..)`.

dmfs:
@lemonboston please rebase

lemonboston:
Rebased

opentaskspal/src/main/java/org/dmfs/opentaskspal/rowsets/Subtasks.java
```diff
@@ -16,29 +16,28 @@
...
 
 /**
  * {@link RowSet} for the subtasks of a given task.
  *
  * @author Gabor Keszthelyi
  */
-public final class Subtasks extends DelegatingRowSet<TaskContract.Tasks>
+public final class Subtasks extends DelegatingRowSet<Tasks>
 {
+    public static final String[] PROJECTION = { Tasks.PARENT_ID };
```
dmfs: What's this `PROJECTION` good for? Isn't it rather pointless? When you're using this you already know that all rows will have the same `PARENT_ID` value, i.e. the value of the parent task `_ID`.
The selection columns don't have to be part of the projection in order to work.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/TaskUri.java
```diff
@@ -0,0 +1,53 @@
...
+import org.dmfs.jems.single.Single;
+import org.dmfs.tasks.contract.TaskContract.Tasks;
+
+
+/**
+ * {@link Single} for the content {@link Uri} that refers to the given {@link RowDataSnapshot}.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class TaskUri implements Single<Uri>
```
dmfs: We definitely need to think of a better way to retrieve the `_ID` or `Uri` of a `RowReference`. There are other use cases in which this is necessary and we don't always have a `RowDataSnapshot` of the row, like when we just inserted it.
We'll keep this for now, but we need to fix that in `ContentPal`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/DueTime.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * {@link Optional} for the due {@link DateTime} of a task.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class DueTime extends DelegatingOptional<DateTime>
+{
+    public static final String[] PROJECTION;
+
+    // TODO Either switch to Iterables (maybe at {@link View#withProjection} as well), or create support for array operations
```
dmfs: I already considered a `Projection` type which can be composed from multiple projections. It might even make sense to move the projection from being a `View` property to being a `RowSet` property which then passes it as a parameter to `View.rows(…)`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/rowsets/Subtasks.java
lemonboston: Right, it was wrong here, `PARENT_ID` is not part of the projection here. Removed it.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/DueTime.java
lemonboston: I can see that that could help a bit, but that wouldn't replace this need to define the required projections as constants on read data classes, right? It's really not nice but since querying and reading the individual columns is independent, I don't see a different solution now. Do you?

About this `Projection` type and possibly changing `View`/`RowSet`, shall we create a ticket in contentpal?
I assume it wouldn't be worth holding this subtask feature up because of that though.

dmfs: FTR: I created this https://github.com/dmfs/ContentPal/issues/134

In general we can't really guarantee that the columns you want to read are present in a `RowDataSnapshot`. Also IMO that's not really the job of the types we have so far.

In order to avoid these static fields you would have to create new types like `Query` and `Entity` which provide the required information to `RowSet` and `View`. IMO that's subject to the next layer on top of ContentPal. It could work like this (sort of):

```java
Iterable<EntitySet<Tasks>> tasks =
    new Query<Tasks>(tasksView,
        new Seq<>(
            new DueEntity(),
            new StartEntity(),
            new TitleEntity())).results();
```

`EntitySet` would have a method to retrieve the value of a specific `Entity`.

The other option would be to always request all fields, that's the only guarantee that the column you need is in the `RowDataSnapshot` you have.

lemonboston: How about doing that for now for subtasks loading, to omit selection and loading all columns?
I suppose the query's speed is the same, to find the rows, and since we won't have many subtasks usually the overhead of loading those extra columns into memory is negligible.

I would remove these constants and the projection from the query.

lemonboston:
Moving this back to Backlog while focusing on bugs. When coming back to this, I suppose I can use `Projection` from latest contentpal for the constants.

dmfs:
@lemonboston this needs to be rebased

lemonboston:
Rebased, squashed, and:
- updated to jems 1.14 to remove `SinglePresent` copy
- replaced `String[]` projection constants with `Projection`
- replaced eager method in `DueTime` with lazy declaration as suggested
- added `@NonNull` annotations to ctor parameters

Ready for check.
Unit tests are yet to be added, I'd like to do them after you checked the classes to avoid potential effort of rewriting tests if something still need to be changed significantly.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/DueTime.java
```diff
@@ -0,0 +1,83 @@
...
+import org.dmfs.rfc5545.Duration;
+import org.dmfs.tasks.contract.TaskContract.Tasks;
+
+
+/**
+ * {@link Optional} for the due {@link DateTime} of a task.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class DueTime extends DelegatingOptional<DateTime>
```
dmfs: please rename to `EffectiveDueTime` otherwise it could be mistaken for the `Due` field. This should also be mentioned in the JavaDoc.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/TaskColor.java
```diff
@@ -0,0 +1,56 @@
...
+import org.dmfs.tasks.contract.TaskContract.Tasks;
+
+
+/**
+ * {@link Color} for a task. Uses {@link Tasks#TASK_COLOR} if available,
+ * fall backs to the mandatory {@link Tasks#LIST_COLOR} if not.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class TaskColor implements Color
```
dmfs: Same here, please call this `EffectiveTaskColor`.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/DueTime.java
```diff
@@ -0,0 +1,83 @@
...
+{
+    public static final Projection<Tasks> PROJECTION = new Composite<>(
+            new MultiProjection<Tasks>(Tasks.DUE, Tasks.DTSTART, Tasks.DURATION),
+            TaskDateTime.PROJECTION);
+
+
+    public DueTime(@NonNull final RowDataSnapshot<Tasks> rowDataSnapshot)
+    {
+        super(new FirstPresent<>(new Seq<>(
```
dmfs: This is still too hard to follow and too complex IMO. Why not just use a very clean and simple declaration like in: https://github.com/dmfs/opentasks/blob/stories/484-taskspal-subtask-data-alt/opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/EffectiveDueDate.java#L40 ?

The `TaskDateTime` implementation in that code is an `Optional` which makes it more versatile and easier to use.

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/TaskDateTime.java
```diff
@@ -0,0 +1,71 @@
...
+public final class TaskDateTime implements Single<DateTime>
+{
+    public static final Projection<Tasks> PROJECTION = new MultiProjection<>(Tasks.TZ, Tasks.IS_ALLDAY);
+
+    private final Long mTimestamp;
+    private final String mTimeZoneId;
+    private final boolean mIsAllDay;
+
+
+    public TaskDateTime(@NonNull Long timestamp, @NonNull String timeZoneId, @NonNull Boolean isAllDay)
```
dmfs: I completely disagree with this class. `TaskDateTime` should be an optional and take a `RowDataSnaphot` and a column name. The current implementation is not task related at all. It just constructs a DateTime from its components. A `TaskDateTime` which deserves the name should act as an adapter from task data to `DateTime`. Since most `DateTime` values of a task are optional this class should implement an `Optional` as well. See https://github.com/dmfs/opentasks/blob/stories/484-taskspal-subtask-data-alt/opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/TaskDateTime.java

opentaskspal/src/main/java/org/dmfs/opentaskspal/readdata/PercentComplete.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author Gabor Keszthelyi
+ */
+public final class PercentComplete extends DelegatingOptional<Integer>
+{
+    public static final Projection<Tasks> PROJECTION = new SingleColProjection<>(Tasks.PERCENT_COMPLETE);
+
+
+    public PercentComplete(@NonNull RowDataSnapshot<Tasks> task)
+    {
+        super(new Mapped<>(new Function<CharSequence, Integer>()
```
dmfs: I've noticed that pretty much all `DelegatingOptional` implementations which return a task value, delegate to a `Mapped` decorator with a Function and `rowDataSnapshot.charData(…)`. I think this could be extracted to a `OptionalTaskCharData` which would be used like so:

```
  super(new OptionalTaskCharData(columName, new Function(){…}));
```
This even works for `TaskTitle` if the function is the identity function.

lemonboston:
You are right, I completely agree with your comments, I've copied the classes from your branch and it's much better with them.
For that `OptionalTaskCharData`, I added `OptionalRowCharData` and `RowCharData` general classes, they could go to ContentPal. Besides them, I've added two more support classes, please check if they make sense. I would open tickets for them in the corresponding libraries.

Ready for check again. (Tests are still yet to come)

dmfs:
Yep, that looks much better. We still need to revisit some of the names. Especially all the `Function` names don't make it clear that they convert from `CharSequence`. Also, we need to check how dmfs/ContentPal#136 could potentially make some of the functions simpler (or maybe even obsolete).

lemonboston:
I've added tests for the non-trivial classes (let me know if you'd like to have tests for the rest as well).

Ready for check.

dmfs:
@lemonboston this needs to be rebases.

dmfs:
you can also get rid of the conversion functions and replace them with method references

lemonboston:
Squashed and rebased on master, and pushed an extra commit with the changes to use the new `RowDataSnapshot` and the method references instead of the `Function`s.
Ready for review.

lemonboston:
@dmfs Do we have any outstanding concern, issue here that I can attend?

dmfs:
I can't remember. I think I was just waiting for 1.1.13 to be released. I'll check this again.

Issue #486, opened 2017-11-20
Travis fails to build PRs


dmfs:
Apparently we have a problem with shallow git clones:
```
fatal: No names found, cannot describe anything.

FAILURE: Build failed with an exception.

* Where:

Build file '/home/travis/build/dmfs/opentasks/build.gradle' line: 13

* What went wrong:

A problem occurred evaluating root project 'opentasks'.

> Process 'command 'git'' finished with non-zero exit value 128
```

dmfs:
Works.

PR #488, opened 2017-11-21
Fix CI build issue of not finding git tag. #486

lemonboston:
As suggested, the `--always` flag is added to the `git describe --tags` command in the gradle file to avoid exiting when no tags found in the shallow clone on the CI.


dmfs:
can you rebase this on https://github.com/dmfs/opentasks/pull/481 so we can see if both travis builds are fixed?

lemonboston:
Rebased on #481 and changed base branch here. Let's see :)

Issue #492, opened 2017-11-21
Create Darkened color decorator to re-use color darkening code


lemonboston:

lemonboston:
Closing this, since this will be implemented in Bolts library.

PR #493, opened 2017-11-21
Darkened color decorator. #492

lemonboston:


lemonboston:
Closing this since this will be implemented in Bolts library.

Issue #396, opened 2017-07-29
"Tasks due" list displays time in 12h format (although 24h format is set in Phone Settings)


alexander-koerschgen:
In the "tasks due" list, the due time for a task is displayed in 12h format (e.g., "Today, 3:00 PM"). However, if I select a task to view its details, the due time is displayed in 24h format ("... 15:00").

The language on my phone is set to "English (United States)", and the "Use 24-hour format" option is active in "Date and time" settings. If I change the latter setting, the due time in detail view is also displayed in 12h format (which, I think, is the correct behavior).

I believe that the app should respect the global 24-hour format setting at all places or provide a custom setting within the app to override it. 

My phone is an Samsung Galaxy S5 Mini (SM-G800F) running Android 6.0.1. I cannot tell for sure, but I think that this issue did not appear on Android 5.

PR #510, opened 2017-11-27
Fix 12/24h format on event list due date-time by using an 5.1.1 Andro…

lemonboston:
…id version of the DateUtils method. #396 

---

Fixed it as discussed on chat.
I just added the old version of the method as a private method of `DateFormatter` as I assume it's best to keep every date-time formatting code there until `Time` is removed form the project. We can refactor after that.

Btw, found the bug on the official tracker: https://issuetracker.google.com/issues/37127319


dmfs:
How about using the workaround only if we have to. I.e. detect whether the device 24h/12h setting is different from the current locale 24h/12h setting and use the original method if it's the same?

It's probably best if we use workarounds only to the extend we have to and stick to the original methods otherwise.

lemonboston:
I've updated it as suggested, so only use old version when locale has AM/PM by default. Ready for check again.

Issue #511, opened 2017-11-27
java.lang.NullPointerException in ColorFieldAdapter


dmfs:
```
java.lang.RuntimeException: 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2684)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2751)
  at android.app.ActivityThread.-wrap12 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1496)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6186)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:889)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:779)
Caused by: java.lang.NullPointerException: 
  at org.dmfs.tasks.model.adapters.ColorFieldAdapter.get (ColorFieldAdapter.java:93)
  at org.dmfs.tasks.EditTaskFragment.onCreateView (EditTaskFragment.java:299)
  at android.support.v4.app.Fragment.performCreateView (Fragment.java:2087)
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1113)
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1295)
  at android.support.v4.app.BackStackRecord.run (BackStackRecord.java:801)
  at android.support.v4.app.FragmentManagerImpl.execPendingActions (FragmentManager.java:1682)
  at android.support.v4.app.FragmentController.execPendingActions (FragmentController.java:388)
  at android.support.v4.app.FragmentActivity.onStart (FragmentActivity.java:607)
  at android.support.v7.app.AppCompatActivity.onStart (AppCompatActivity.java:181)
  at android.app.Instrumentation.callActivityOnStart (Instrumentation.java:1249)
  at android.app.Activity.performStart (Activity.java:6701)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2647)
```

lemonboston:
So according to this: https://github.com/dmfs/opentasks/issues/521#issuecomment-348309161 ,
I assume this bug may be caused by opening `EditTaskFragment` for a deleted / non-existent task.
[...]

PR #514, opened 2017-11-28
Fix an NPE by adding a null-check on ContentSet in EditTaskFragment o…

lemonboston:
…nCreateView. #511 

---

I simply added a null check around the point where the NPE came from.
The logic in this `EditTaskFragment` is rather tangled, so please check yourself too that the change makes sense and fits the logic. I also don't know how to reproduce the crash.

(I expect that if/when we replace `ContentSet` and the loading with RxJava, then this `Fragment` would be simplified significantly, so I assume it's not worth trying to improve on the current code and solution.)

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
@@ -296,9 +296,13 @@ public class EditTaskFragment extends SupportFragment implements LoaderManager.L
                         // ensure we're using the latest values
                         mValues.update(mAppContext, CONTENT_VALUE_MAPPER);
                     }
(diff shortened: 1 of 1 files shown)


dmfs:
Not sure how `mValue` can be `null` at that position. This could be a life-cycle issue.
Just ignoring the `null` pointer could cause a crash later on when `mValues` is used again.

I think the life-cycle handling needs to be fixed. I'll check that.

lemonboston:
Moved to Icebox (see https://github.com/dmfs/opentasks/issues/511#issuecomment-348331341)

dmfs:
Closing this. We're not simply ignoring the null condition. If this is still an issue after #521 was fixed, I'd like to know about it.

Issue #382, opened 2017-07-11
Google Play Crash Report: android.database.sqlite.SQLiteException


dmfs:
```
android.database.sqlite.SQLiteException: 
  at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method:0)
  at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1113)
  at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:686)
  at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
  at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
  at android.database.sqlite.SQLiteStatement.<init>(SQLiteStatement.java:31)
  at android.database.sqlite.SQLiteDatabase.executeSql(SQLiteDatabase.java:1769)
  at android.database.sqlite.SQLiteDatabase.execSQL(SQLiteDatabase.java:1700)
  at org.dmfs.provider.tasks.FTSDatabaseHelper.onUpgrade(FTSDatabaseHelper.java:187)
  at org.dmfs.provider.tasks.TaskDatabaseHelper.onUpgrade(TaskDatabaseHelper.java:783)
  at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:257)
  at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:164)
  at org.dmfs.provider.tasks.SQLiteContentProvider.update(SQLiteContentProvider.java:191)
  at org.dmfs.provider.tasks.TaskProvider.update(TaskProvider.java:104)
  at android.content.ContentProvider$Transport.update(ContentProvider.java:235)
  at android.content.ContentResolver.update(ContentResolver.java:995)
  at org.dmfs.provider.tasks.ContentOperation.fire(ContentOperation.java:342)
  at org.dmfs.provider.tasks.TaskProvider$1.run(TaskProvider.java:1152)
  at android.os.Handler.handleCallback(Handler.java:725)
  at android.os.Handler.dispatchMessage(Handler.java:92)
  at android.os.Looper.loop(Looper.java:175)
  at android.os.HandlerThread.run(HandlerThread.java:60)
```

No further information.

dmfs:
@lemonboston please look into this and let me know if you can solve it.

lemonboston:
[...]

PR #515, opened 2017-11-28
Fix crashes caused by CREATE INDEX SQL commands upon DB update by add…

lemonboston:
…ing IF NOT EXISTS clause. #382 

---
Steps to reproduce the crash:
- clean install from master with `TaskDatabaseHelper#DATABASE_VERSION = 15`
(may also need `allowBackup=false` in manifest)
- add tasks
- install update with version `TaskDatabaseHelper#DATABASE_VERSION = 16`

Doing the same with this branch should not result in a crash.

### DIFF
--- a/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
+++ b/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
@@ -533,7 +533,7 @@ public class TaskDatabaseHelper extends SQLiteOpenHelper
         {
             buffer.append(" UNIQUE ");
         }
-        buffer.append("INDEX ");
+        buffer.append("INDEX IF NOT EXISTS ");
         buffer.append(table).append("_").append(fields[0]).append("_idx ON ");
         buffer.append(table).append(" (");
         buffer.append(fields[0]);


opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
```diff
@@ -531,9 +531,9 @@ public final static String createIndexString(String table, boolean unique, Strin
         buffer.append("CREATE ");
         if (unique)
         {
-            buffer.append(" UNIQUE ");
+            buffer.append("UNIQUE ");
         }
-        buffer.append("INDEX ");
+        buffer.append("INDEX IF NOT EXISTS");
```
dmfs: How can this work? This doesn't leave a space between "EXISTS" and the table name.

lemonboston: Sorry, the commited version missed the space, the 'IF NOT EXIST ' append was in a separate line when I tested.
This is why the build failed as well. Fixed now.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
```diff
@@ -531,10 +531,10 @@ public final static String createIndexString(String table, boolean unique, Strin
         buffer.append("CREATE ");
         if (unique)
         {
-            buffer.append(" UNIQUE ");
+            buffer.append("UNIQUE ");
         }
-        buffer.append("INDEX ");
-        buffer.append(table).append("_").append(fields[0]).append("_idx ON ");
+		buffer.append("INDEX IF NOT EXISTS ");
+		buffer.append(table).append("_").append(fields[0]).append("_idx ON ");
```
dmfs: why the indentation now?

lemonboston: I've squashed-rebased now to fix this, please check again. 

But finally it turned out what these look like identation changes are: the file has tabs instead of spaces and the IDE keeps them even though the code style uses spaces.

There is this settings:
screenshot from 2017-11-29 14-39-18

And this notification:
screenshot from 2017-11-29 14-41-20
(So in my commit those 2 lines were reformatted for some reason.)

I guess we could fix this in a separate ticket. Turn off that settings and re-format alll project files, to remove the tabs.

Issue #501, opened 2017-11-23
caption of cancel button has line wrap


voxspox:
The caption of the cancel button in the synchronized lists view wraps to "cance" / "l". This happens for a S4 mini with LineageOS 14.1 (Android 7.1.2)

dmfs:
Note, this view is going to be removed. We'll change how and where the visible task lists are selected.

Anyway, can you attach a screen shot?

voxspox:
![screenshot_20171124-001045](https://user-images.githubusercontent.com/1503565/33190466-530e097e-d0ad-11e7-8fef-57d9a05c0d80.png)

dmfs:
Shouldn't be hard to fix that.

PR #519, opened 2017-11-29
Fix the wrapping of "CANCEL" on settings button on Android 7.x. #501

lemonboston:
Note that the text wrapping doesn't happen on Android 6.0, but I could reproduce it on 7.0 (may need to enter a longer text in place on 'cancel' to see it)

### DIFF
--- a/opentasks/src/main/res/layout/fragment_synced_task_list.xml
+++ b/opentasks/src/main/res/layout/fragment_synced_task_list.xml
@@ -4,7 +4,7 @@
         android:layout_width="match_parent"
         android:layout_height="fill_parent"
         android:orientation="vertical"
-        tools:context=".SettingsListFragment">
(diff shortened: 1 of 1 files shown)


dmfs:
This seems to break the tab bar on tablets for me, see this picture:
screenshot_1512055827

dmfs:
I saw that with and without rebasing on the current master.

lemonboston:
Just to confirm, the issue on your screenshot is the invisible tab bar icons, right?
I can't reproduce it and I think it would be very strange if this unrelated layout change caused anything like that. Wasn't it caused by some faulty Instant Run build? That's the only thing I can think of...
But if it is consistent, could you please tell me the steps how I can reproduce this possibly starting from a clean install?

dmfs:
right. Not sure what the problem was. I can't reproduce it right now. Maybe it was a built error.

Issue #353, opened 2017-07-03
App bar color of empty fragment not updated 


dmfs:
The app bar color of the empty fragment should match the current color of the list app bar. See the attached screen shot.
![device-2017-07-03-151555](https://user-images.githubusercontent.com/1614987/27794793-e5fa3af2-6003-11e7-8314-347e9b14925b.png)

PR #520, opened 2017-11-29
Fix app bar colors when empty task fragment is shown in two pane mode…

lemonboston:
…. #353

---

I did it the other way around, so set the list app bar color to the primary color when the empty fragment is shown, since no task(list) is selected that time. This is easier to implement and also safer I think, instead of somehow keeping the color of the last selected task.


lemonboston:
I guess it's better to use primary color from the theme attribute.... I am going to update it.

lemonboston:
Imported `bolts-color` and use `AttributeColor(R.attr.colorPrimary)` now for the app bar color.
Ready for check.

opentasks/src/main/java/org/dmfs/tasks/ViewTaskFragment.java
```diff
@@ -175,23 +176,23 @@ public void run()
...
+        void onDelete(Uri taskUri);
 
         /**
          * Notifies the listener about the list color of the current task.
          *
          * @param color
          *         The color.
          */
-        public void updateColor(int color);
+        void updateColor(@ColorInt int color);
```
dmfs: How about changing this to take a `Color` instead?

lemonboston: I've changed it to use `Color`, also created https://github.com/dmfs/opentasks/issues/522 to use `Toned` as next step.
Ready for check.

Issue #457, opened 2017-10-17
Revisit Relations Properties Contract


dmfs:
The relations properties need to be revisited before put in production. Some values might be obsolete.

dmfs:
Also consider adding a column to identify reverse relations or automatically created relations. The idea is to automatically create a child->parent relationship for every parent->child relationship (and vice versa). Though there should be something which that identifies these automatic relations so a sync-adapter can filter them out.

PR #524, opened 2017-11-30
Revisit Relation contract. Implements #457

dmfs:
* remove all the `public static final` modifiers
* remove RELATED_URL
* remove some currently unsupported RelTypes
* convert RelTypes from enum to ints to allow unsupported reltypes.

### DIFF
--- a/opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java
+++ b/opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java
@@ -22,7 +22,6 @@ import android.database.sqlite.SQLiteDatabase;
 
 import org.dmfs.provider.tasks.TaskDatabaseHelper;
 import org.dmfs.tasks.contract.TaskContract.Property.Relation;
-import org.dmfs.tasks.contract.TaskContract.Property.Relation.RelType;
 import org.dmfs.tasks.contract.TaskContract.Tasks;
 
 
@@ -44,21 +43,13 @@ public class RelationHandler extends PropertyHandler
 
         Long id = values.getAsLong(Relation.RELATED_ID);
         String uid = values.getAsString(Relation.RELATED_UID);
-        String uri = values.getAsString(Relation.RELATED_URI);
 
-        if (id == null && uri == null && uid != null)
+        if (id == null && uid != null)
         {
             values.putNull(Relation.RELATED_ID);
-            values.putNull(Relation.RELATED_URI);
         }
-        else if (id == null && uid == null && uri != null)
+        else if (id != null && uid == null)
         {
-            values.putNull(Relation.RELATED_ID);
-            values.putNull(Relation.RELATED_UID);
-        }
-        else if (id != null && uid == null && uri == null)
-        {
-            values.putNull(Relation.RELATED_URI);
             values.putNull(Relation.RELATED_UID);
         }
         else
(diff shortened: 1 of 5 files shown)


dmfs:
@lemonboston I want to handle the "reverse relation" issue in a separate story

opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java
```diff
@@ -44,21 +43,13 @@ public ContentValues validateValues(SQLiteDatabase db, long taskId, long propert
 
         Long id = values.getAsLong(Relation.RELATED_ID);
         String uid = values.getAsString(Relation.RELATED_UID);
-        String uri = values.getAsString(Relation.RELATED_URI);
 
-        if (id == null && uri == null && uid != null)
+        if (id == null && uid != null)
```
lemonboston: Shouldn't this `validate()` method just simply throw if none or both of `RELATED_ID` and `RELATED_UID` are present?
Isn't the value already `null` when `values.putNull(Relation.RELATED_ID);` is called?

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/RelationData.java
```diff
@@ -37,23 +36,23 @@
 public final class RelationData extends DelegatingRowData<TaskContract.Properties>
 {
     public RelationData(@NonNull RowSnapshot<TaskContract.Tasks> relatingTask,
-                        @NonNull RelType relType,
+                        @NonNull int relType,
```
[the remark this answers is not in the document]
dmfs: the `@NonNull` should be removed. I'll take care of that.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java
dmfs: that would only be the case for new rows. When you override a row we need to reset the other value explicitly. That being said, I already considered making rows write-once. So you can only create and delete them. Updates would throw an exception. I'm starting to get used to this idea and it would certainly make a couple of things much simpler.

dmfs:
@lemonboston ready for re-review.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/handler/RelationHandler.java
lemonboston: I still don't get this - so just by looking at this code:
```java
Long id = values.getAsLong(Relation.RELATED_ID);
String uid = values.getAsString(Relation.RELATED_UID);

if (id == null && uid != null)
{
    values.putNull(Relation.RELATED_ID);
}
```
and just ignore the 'uid' for a second and inline the 'id' variable, it becomes this:
```java
if (values.getAsLong(Relation.RELATED_ID) == null)
{
    values.putNull(Relation.RELATED_ID);
}
```
So if the value is `null` we set it to `null`..   Or do I miss something?

dmfs: `getAsLong` also returns `null` if the value is absent (i.e. not not present in the `ContentValues`), so we need to explicitly put a `null` value to ensure we override the old value, otherwise it would remain unchanged.

lemonboston: oh, I see, thanks

Issue #531, opened 2017-12-02
TransactionEndTasks


dmfs:
In order to make changes to the task database after all operations had been executed but still before the transaction is finished, I want to provide one or more `TransactionEndTask`s which are executed right before the transaction is about to be finished successfully.
---

PR #532, opened 2017-12-02
Introduce TransactionEndTask. Implements #531

dmfs:
This commit introduces a TransactionEndTask which is run right before a transaction is ended.
It also adds a simple implementation of TransactionEndTask which performs the call to setTransactionSuccessful() and runs it as the very last of these tasks.


lemonboston:
The changes look correct to me.

But looking into these transaction codes and seeing the javadoc for `beginTransaction` about the suggested usage:
```java
db.beginTransaction();
try {
   ...
   db.setTransactionSuccessful();
} finally {
   db.endTransaction();
}
```
this clearly looks like a template, which is in fact repeated in `SQLiteContentProvider`. I suppose it could be extracted, and having a type that is executed in the middle, taking `SQLiteDatabase` as parameter.
The end transaction could be of the same type.
Not sure if this is worth attending now or not, so I just set my review to approved.

lemonboston:
Something like this (quickly written down, not sure if it's correct):
```java
interface SqlLiteTransactionContent<T>
{
       T execute(SqlLiteDatabase db);
}
// T could be Uri or Integer(count)

interface SqlLiteTransaction
{
       void execute(SqlLiteTransactionContent content) 
}

// end transaction could be SqlLiteTransactionContent<T> decorator, returning the delegate result

// usage:
new BasicSqlLiteTransaction(db).execute(new Ended(endContents, new Insert(..)));
// the end transactions could also be added by secondary SqlLiteTransaction ctors
// the template would be implemented in BasicSqlLiteTransaction (it could even be SqlLiteTransactionContent decorator, although not sure if that's a good idea)
```

dmfs:
Actually I already have something like that in a separate branch. It's slightly more complicated than that because as you can see in the code it depends on a ThreadLocal variable whether the current transaction is marked successful right away or not. I'm currently in the process of refactoring this class (which originated in Android's Calendar Provider).

The major issue is that `applyBatch` receives `ContentProviderOperation`s which call the provider but need to be executed in a single transaction, hence we can't just commit each transaction in `insert`, etc.

Anyway, I want to do this step by step. The call to `setTransactionSuccessful` will certainly be moved somewhere else during that process.

lemonboston:
I see. This redesign and refactor certainly looks like something that may result in a big changeset.

PR #536, opened 2017-12-04
Upgrade build tools & Gradle Plugin. Fixes #534

### LINKED ISSUE #534: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR)
```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 0 >>> org.dmfs.tasks <<<
backtrace:
  #00  pc 0000000000034b58  /system/lib/libc.so (dlfree+799)
  #01  pc 0000000000015c69  /system/lib/libandroidfw.so (_ZN7android13ResStringPool6uninitEv+36)
  #02  pc 0000000000016bef  /system/lib/libandroidfw.so (_ZN7android10ResXMLTree6uninitEv+14)
  #03  pc 0000000000016bc9  /system/lib/libandroidfw.so (_ZN7android10ResXMLTreeD1Ev+4)
  #04  pc 0000000000011dab  /system/lib/libandroidfw.so (_ZN7android12AssetManager10getPkgNameEPKc+274)
  #05  pc 0000000000011e23  /system/lib/libandroidfw.so (_ZN7android12AssetManager18getBasePackageNameEj+54)
  #06  pc 000000000008a0c7  /system/lib/libandroid_runtime.so
  #07  pc 00000000732f2ca1  /data/dalvik-cache/arm/system@framework@boot.oat
```

dmfs:


lemonboston:
I get this Lint error when built locally:
image
Can be fixed with adding the exclude to bolts:
```
    implementation ('com.github.dmfs.bolts:color-bolts:' + BOLTS_VERSION) {
        exclude group: 'com.android.support'
    }
```
This is unrelated to this issue, but I suppose we should fix this, too, either here, or in a separate ticket.

dmfs:
I've upgraded the support lib as well.

lemonboston:
Looks good to me.

We should still exclude support libs from bolts dependency though, since the error will come back when we update bolts with higher support lib version. Although, I thought this problem would disappear with using `implementation` config, but apparently it hasn't. (I've checked, the used bolts version already uses `implementation` for the support lib.)

PR #549, opened 2017-12-05
Update minSdkLevel to 15. Remove all obsolete code and resources. #392

lemonboston:
A lot of code and resources could be removed.


opentasks/src/main/java/org/dmfs/tasks/utils/BaseActivity.java
```diff
@@ -83,24 +81,15 @@ protected void onSaveInstanceState(Bundle outState)
...
     {
         super.onPause();
-        /*
-         * On older SDK version we have to store permanent data in onPause(), because there is no guarantee that onStop() will be called.
-		 */
-        if (VERSION.SDK_INT < VERSION_CODES.HONEYCOMB)
-        {
-            RetentionMagic.persist(this, mPrefs);
-        }
+        RetentionMagic.persist(this, mPrefs);
```
dmfs: This should be removed. It was only enabled on SDK<11.

dmfs:
@lemonboston please rebase

opentasks/src/main/java/org/dmfs/tasks/utils/BaseActivity.java
lemonboston: Good catch, thanks, I've corrected it.

Squashed and rebased, ready for check.

lemonboston:
Yep, this is the best :)

Issue #533, opened 2017-12-02
Refactor EntityProcessors


dmfs:
EntityProcessors are currently interfaces with 6 methods and they are stored in a List. This can be improved by using the decorator pattern.

PR #551, opened 2017-12-06
Refactor entity processor chain. Implements #533

dmfs:
This is the first step of a major refactoring of the provider. This step essentially merges the "before*" and "after*" methods of `EntityProcessor` and converts most implementations into decorators.


opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/Logging.java
```diff
@@ -0,0 +1,66 @@
...
+    public Logging(EntityProcessor<T> delegate)
+    {
+        mDelegate = delegate;
+    }
+
+
+    @Override
+    public T insert(SQLiteDatabase db, T entityAdapter, boolean isSyncAdapter)
+    {
+        Log.d("Logging EntityProcessor", "before insert");
```
lemonboston: Move the tag to a constant?

Issue #537, opened 2017-12-04
Move instance generation to a TransactionEndTask


dmfs:
In order to simplify instance generation for recurring tasks, I want a `TransactionEndTask` which (re-) creates instances of new and updated tasks.
---
At present we have a dedicated `TaskInstancesProcessor` class which takes care of this while a task is being inserted. This is handy for single tasks, but it makes it more complex for recurring tasks, especially when overrides exist.

The idea is to flag new tasks and tasks with updates that affect the instances table and process them after a batch has been executed. This should happen within the same transaction to ensure consistency.

dmfs:
We've decided to keep the instance creation in the processor chain.

PR #554, opened 2017-12-06
Move instance processing to TransactionEndTask. Implements #537

dmfs:
TaskInstancesProcessor now only sets an instance_stale flag when appropriate. Instances are created after the batch has been processed.
UpdateInstancesTask doesn't handle recurrence yet. That's a different story.

### DIFF
--- a/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
+++ b/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
@@ -22,6 +22,7 @@ import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
 import android.util.Log;
 
+import org.dmfs.provider.tasks.model.TaskAdapter;
 import org.dmfs.tasks.contract.TaskContract;
 import org.dmfs.tasks.contract.TaskContract.Properties;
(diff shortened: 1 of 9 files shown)


opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
```diff
@@ -783,6 +786,12 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion)
             db.execSQL(createIndexString(Tables.INSTANCES, false, TaskContract.Instances.INSTANCE_START_SORTING));
             db.execSQL(createIndexString(Tables.INSTANCES, false, TaskContract.Instances.INSTANCE_DUE_SORTING));
         }
+        if (oldVersion < 17)
+        {
+            // Note: in general only a small fraction of all rows will have this flag set, so an index can improve performance on large tables
+            db.execSQL("alter table " + Tables.TASKS + " add column " + Tasks.INSTANCES_STALE + " INTEGER DEFAULT 1;");
+            db.execSQL(createIndexString(Tables.TASKS, false, Tasks.INSTANCES_STALE));
```
lemonboston: Maybe run `UpdateInstancesTask` right away here to update all existing instances?

dmfs:
@lemonboston this is ready for re-review.

dmfs:
right, I have a better idea

PR #558, opened 2017-12-07
Added Chinese(Simplified) translation

vellxn:


dmfs:
@lemonboston please review.

dmfs:
@lemonboston how about you?

dmfs:
@vellxn can you also provide translations for these plurals? https://github.com/dmfs/opentasks/blob/a6c5e6a0be3aef2fbee872d0078443d34989695b/opentasks/src/main/res/values/strings.xml#L28-L31

Issue #522, opened 2017-11-30
Use Toned(Color) from Bolts library to replace dynamic color adjusting code


lemonboston:
Worth searching for all occurences of `colorToHSV` usage in the project and replace those methods with declarative `Color` classes.

PR #560, opened 2017-12-08
Extract dynamic color adjusting codes to Color classes. #522

lemonboston:


opentasks/src/main/java/org/dmfs/tasks/utils/colors/AdjustedForFab.java
```diff
@@ -0,0 +1,66 @@
...
+    public AdjustedForFab(@ColorInt int listColor)
+    {
+        this(new ValueColor(listColor));
+    }
+
+
+    @Override
+    public int argb()
+    {
+        // TODO Use HSL instead?
```
dmfs: this should be a `DelegatingColor` delegating to `Toned`. Since `Toned` operates on HSL you can use this simplified function:
```java
lightness>0.5 ? lightness/1.2 : lightness*1.2
```

opentasks/src/main/java/org/dmfs/tasks/model/adapters/ColorFieldAdapter.java
```diff
@@ -90,31 +90,21 @@ public ColorFieldAdapter(String fieldName, Integer defaultValue, float darkenThr
...
-        return darkenColor(super.get(values), mDarkenThreshold);
+        return new SmartDarkening(mDarkenThreshold, super.get(values)).argb();
     }
 
 
     @Override
     public Integer get(Cursor cursor)
     {
-        return darkenColor(super.get(cursor), mDarkenThreshold);
+        return new SmartDarkening(mDarkenThreshold, super.get(cursor)).argb();
```
dmfs: Did you consider to make this return `Color` instead of `Integer`?

lemonboston: No, I didn't, since there is plan to remove `ContentSet` and current loading mechanisms, I thought it's probably not worth spending time with those structures.
Let me know if you think it's worth it. I would like to do that in a separate ticket though.

opentasks/src/main/java/org/dmfs/tasks/utils/colors/AdjustedForFab.java
lemonboston: Ok, that's much better. I've updated it.

lemonboston:
Ready for check again.

opentasks/src/main/java/org/dmfs/tasks/model/adapters/ColorFieldAdapter.java
dmfs: Ok, let's do that in a separate ticket.

lemonboston: Ticket added: https://github.com/dmfs/opentasks/issues/561

dmfs:
This doesn't work yet. This is how it looks like with a very light list color: 
screenshot_1512746242
That should have been darkened.

dmfs:
@lemonboston dmfs/Bolts#17 has been merged

lemonboston:
I've updated this pull request so that none of the color adjusting calculations are changed in effect, so this is only about extracting them to `Color` classes. Since we ran into issues, questions with the color encodings, I assumed it's better to address any actual change in separate ticket.

Squashed, rebased, ready for review.

dmfs:
@lemonboston please resolve conflicts

lemonboston:
Rebased.

opentasks/src/main/java/org/dmfs/tasks/utils/colors/Blended.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.dmfs.android.bolts.color.Color;
+
+
+/**
+ * Color blended from two others. See implementation for details.
+ *
+ * @author Gabor Keszthelyi
+ */
+public final class Blended implements Color
```
dmfs: Note, the purpose of this code is essentially the reverse of `Mixed`. So if `Mixed` blends the colors `A` and `B` into `C`, this class returns `B` when you give it `A` (`baseColor`), `C` (`targetColor`) and the alpha value of `B`. We use this in the editor to determine the color of the action bar, so it mixes with the background in a way that keeps the result the same (while the alpha of the action bar goes towards opaque)

lemonboston: Ok, I've added this note as javadoc to the class.

Issue #564, opened 2017-12-11
Refactor Instance creattion and add `INSTANCE_ORIGINAL_TIME`


dmfs:
In order to prepare for recurrence support, we need to refactor instance creation. In particular we need to introduce an `INSTANCE_ORIGINAL_TIME` column. This column serves as sort of a key for instances of a specific task. We can't use `INSTANCE_START` because this might be `null` and is not `unique`. Same goes for 'INSTANCE_START_SORTING` and the due columns.

PR #565, opened 2017-12-11
Add INSTANCE_ORIGINAL_TIME column to instances. Implements #564

dmfs:
This adds a column to the instances table to hold the original date and time of an instance. To make this easier instance creation has been refactored into smaller components.
This is a prerequisite for recurrence support.


opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/InstanceDateTimeData.java
```diff
@@ -0,0 +1,111 @@
...
+
+
+    private void putDateValue(ContentValues values, Optional<DateTime> value, String timestampColumn, String sortingColumn, TimeZone localTimeZone)
+    {
+        if (value.isPresent())
+        {
+            // add timestamp and sorting
+            DateTime dateTime = value.value();
+            values.put(timestampColumn, dateTime.getTimestamp());
+            values.put(sortingColumn, dateTime.isAllDay() ? dateTime.getInstance() : dateTime.shiftTimeZone(localTimeZone).getInstance());
```
lemonboston: I see here that the sorting column has different value than the timestamp one..   could you help explain a bit what is this column for? Is it related to performance optimization?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/tasks/Instantiating.java
```diff
@@ -45,10 +47,14 @@
     /**
      * This is a field adapter for a pseudo column to indicate that the instances may need an update, even if no relevant value has changed. This is useful to
      * force an update of the sorting values when the local timezone has been changed.
+     * <p>
+     * TODO: get rid of it
      */
     private final static BooleanFieldAdapter<TaskAdapter> UPDATE_REQUESTED = new BooleanFieldAdapter<TaskAdapter>(
             "org.dmfs.tasks.TaskInstanceProcessor.UPDATE_REQUESTED");
 
+    private final static int INSTANCE_COUNT_LIMIT = 1000;
```
lemonboston: Is this number about how many recurring instance will be created? If yes, isn't this way too much? I mean that I, as a user, would be fine with having let's say the next 30 occurrence available for me in the app.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -199,7 +232,111 @@ public void testInsertTaskWithStartAndDue()
...
+                ))
+        ));
+    }
+
+
+    /**
+     * Create task with start and due and update it with new values, check datetime values including generated duration.
+     */
+    @Test
+    public void testInsertTaskWithStartAndDueMovedForward()
```
lemonboston: About 2/3 of the tests fail for me, including this one, with:
`java.lang.AssertionError: android.content.OperationApplicationException: wrong number of rows: 0`

Do they pass for you?

opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
```diff
@@ -1035,6 +1035,14 @@ public static Uri getSearchUri(String authority, String query)
          */
         String INSTANCE_DURATION = "instance_duration";
 
+        /**
+         * The start of the original instance as specified in the master task. For non-recurring task instances this equals the value of {@link
+         * #INSTANCE_START}, except that `null` values are represented as `0`.
+         * <p>
+         * For recurring tasks, these are the timestamps which have been derived from the recurrence rule or dates, except those specified as exdates.
```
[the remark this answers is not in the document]
dmfs: `exdates` is not a typo. It's iCalendar terminology for `Exclusion dates`, i.e. dates at which a recurring event does not occur even if the rule says so.

Using `0` instead of `null` makes it much easier to handle, because we can just compare the values without doing any `null` checks. Semantically it doesn't really matter, because recurring events can't have `null` here, by definition.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -199,7 +232,111 @@ public void testInsertTaskWithStartAndDue()
                 new Assert<>(task, new TimeData(start, due)),
                 new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task, new AllOf(
                         new EqArg(Instances.INSTANCE_START, start.getTimestamp()),
+                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, start.getTimestamp()),
+                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
+                        new EqArg(Instances.INSTANCE_START_SORTING, start.getInstance()),
```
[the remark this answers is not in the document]
dmfs: `getInstance()` returns the instance date and time values (i.e. year, month, day, hour, minutes and seconds) packed into a single long. These vales are represented in the time zone of the event (or local time zone for floating events). In contrast to `getTimestamp` this allows correct sorting if all-day and non-all-day events are mixed. The `*_SORTING` columns are not meant to be evaluated other than using them for sorting. There is no guarantee about the "format" of this long value and it might change with every new version.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: They pass for me and for Travis. If they don't pass for you, there might be an issue with the database migration code. It's either not executed or incorrect. I'll check that.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/tasks/Instantiating.java
dmfs: That depends, if you have a daily recurring task 30 might not be sufficient. AFAIK, Android's calendar provider has a limit of 2000. Let's check if that becomes a bottleneck. The iterator is quite fast and can do that in just a few ms. Instance creation might take a little longer though. I also was planning to add a date limit (i.e. don't expand further than 10 years from now). But that's merely an optimization. The good thing about how this is done is, we can change these limits later on without breaking anything.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/InstanceDateTimeData.java
dmfs: As explained above, this is to guarantee correct sorting of floating events. Since they are stored with a UTC timestamp, they would show up at wrong dates and times when sorted by timestamp along with non-floating tasks.
I.e. consider a floating task on June 1st at 8:00 "local time" and another one on June 1st at 9:00 CET. The floating task is stored with a UTC timestamp of "June 1st, 8:00" the CET task is stored with a UTC timestamp of "June 1st 7:00". So if you're located in CET and you sort these by timestamp the 8:00 task will be shown after the 9:00 task.

The sorting columns hold a representation in "local time". So the floating task is stored as "June 1st 8:00" and the absolute task is stored as "June 1st 9:00". Sorting them will show them in the correct order.

Note: when you change the time zone of your device, we need to re-calculate the sorting columns of all tasks and instances. That's why we have this `UPDATE_REQUESTED` hack in the `Instantiating` processor. It ensures all instances are updated even if no datetime value had been modified.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: btw, If you uninstall the test project from your device, the tests should pass on the next launch. This would clearly indicate a migration issue.

dmfs:
@lemonboston I've fixed the tests. Please check again.

dmfs:
That's odd. The `-timezone Europe/Paris` parameter doesn't seem to work on Travis :-|
```
qemu-system-armel: -timezone: invalid option
```

dmfs:
I removed the `timezone` argument for now to make it work.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -230,12 +369,51 @@ public void testInsertWithStartAndDuration()
...
+
+        assertThat(new Seq<>(
+                new Put<>(taskList, new EmptyRowData<TaskLists>()),
+                new Put<>(task, new TimeData(start, duration)),
+                // update the task with a the same start in a different time zone
+                new Put<>(task, new TimeData(startNew, duration))
+
+        ), resultsIn(mClient,
+                new Assert<>(task, new TimeData(startNew, duration)),
+                // note that, apart from the time zone, all values stay the same
```
lemonboston: Why is it that the values apart from timezone don't have to be updated in the Instance table?

dmfs: because the instances table contains only timestamps (which are always in UTC, so they don't change when you shift the event to a different time zone keeping the absolute time constant) and sorting values which only depend on the local time zone of the device and won't change neither as along as you keep the absolute time constant.

dmfs: Also note that the time zone is not even stored in the instances table and there is no need to test it separately because it's already covered by the previous assertion. So if you just shift a task to a different time zone keeping the absolute time constant, the instances table won't change at all.

opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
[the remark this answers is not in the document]
lemonboston: ok, thanks

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
lemonboston: Tests pass now after the fix, on non-UTC timezone device, too.

Created ticket to try to avoid this in the future: https://github.com/dmfs/opentasks/issues/571

opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/InstanceDateTimeData.java
lemonboston: I see, thanks for explaining.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
lemonboston: Thanks, I understand it now from your other comment, too.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/tasks/Instantiating.java
```diff
@@ -75,141 +87,143 @@ public Instantiating(EntityProcessor<TaskAdapter> delegate)
...
+                {
+                    // there is no new instance for this old one, remove it
+                    existingInstances.moveToPosition(next.right().value());
+                    db.delete(TaskDatabaseHelper.Tables.INSTANCES,
+                            String.format(Locale.ENGLISH, "%s = %d", TaskContract.Instances._ID, existingInstances.getLong(idIdx)), null);
+                }
+                else if (!next.right().isPresent())
+                {
+                    // there is no old instance for this new one, add it
+                    ContentValues values = next.left().value();
```
lemonboston: I wanted to check the coverage here in the IDE, to see if `TaskProviderTest` covers this, but since it's an android test, it is not possible of course. I've checked in a different project that Roboelectric tests can be run with coverage. So I am wondering whether it would be useful to add a Roboelectric version as well of this test.
Android test is still desirable and the more important one, I think, because of reasons mentioned earlier, but it wouldn't be a big effort to add the same test (with code-reuse) as a roboelectric one as well, to be able to see code coverage of these rather complex codes in the provider.
(It probably requires a shared source set between test and androidTest.)
What do you think?

dmfs: I'm wondering if there isn't a way to get the coverage of Android tests.
I wouldn't want to duplicate the tests for Robolectric.

Also, we have much more options to test all the provider classes in isolation using Robolectric. Most of them (at least the newer ones) should be quite testable.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/Limited.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.dmfs.provider.tasks.utils;
+
+import java.util.Iterator;
+
+
+/**
+ * An {@link Iterable} which limits the number of elements.
+ * <p>
+ * TODO: move to jems
```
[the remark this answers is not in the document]
dmfs: done https://github.com/dmfs/jems/issues/132

opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/LimitedIterator.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.dmfs.iterators.AbstractBaseIterator;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+
+/**
+ * An {@link Iterator} which limits the number elements.
+ * TODO: move to jems
```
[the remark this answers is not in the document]
dmfs: covered by https://github.com/dmfs/jems/issues/132

opentasks-provider/src/main/java/org/dmfs/provider/tasks/utils/Range.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Iterator;
+
+
+/**
+ * An {@link Iterable} which iterates a range of numbers.
+ *
+ * @author Marten Gajda
+ */
+public final class Range implements Iterable<Integer>
```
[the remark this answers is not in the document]
dmfs: https://github.com/dmfs/jems/issues/130

dmfs:
Ok let's see how this works out.

Issue #393, opened 2017-07-26
Remove CHANGELOG file.


dmfs:
The CHANGELOG file is abandoned and can be removed after migrating the change log to https://github.com/dmfs/opentasks/releases

dmfs:
The content of the ChangeLog file has been made available at https://github.com/dmfs/opentasks/releases

PR #573, opened 2017-12-12
remove changelog and redirect to Github releases. Implements #393

dmfs:


dmfs:
@lemonboston this should be an easy one :-)

Issue #562, opened 2017-12-09
"OpenTasks" wurde beendet


Diet-mar:
I still have the issue that OpenTasks terminates when I try to add a task; be it via the small + or the big + in the red circule.
OpenTasks version. 1.1.11 from F-Droid
Android 6 without Google stuff: Fairphone Open
The task list is on a self hosted radicale server. Synchronisation works through DAVDroid.
I already created a fresh ics file without change, cleared cache and dalvik of my mobile phone.
I cannot find the location where the error messages should be stored.

dmfs:
There should be a crash message in the system log which can be read with `adb` and (if root access is available) various logger apps. We can't reproduce this error, so without a stack trace we can't really do very much other than guesswork. We're currently checking the crash reports on Google Play as well.

At what point does the app crash? Already when pressing the `+` button or when you hit the save button?

Diet-mar:
Thanks dmfs for the quick response.

The crash happens when I click the save button. Independant of what kind of data I enter.
OpenTasks was running, then I start the log until I have clicked the OK of the crash message.
There is a warning
> ContentResolver: Failed to get type for: content://org.dmfs.tasks (Unsupported URI: content://org.dmfs.tasks)

And lateron the fatal exception:
> 12-12 21:08:06.764 10017 10017 E AndroidRuntime: FATAL EXCEPTION: main
12-12 21:08:06.764 10017 10017 E AndroidRuntime: Process: org.dmfs.tasks, PID: 10017
12-12 21:08:06.764 10017 10017 E AndroidRuntime: java.lang.NumberFormatException: Invalid long: "null"
12-12 21:08:06.764 10017 10017 E AndroidRuntime: 	at java.lang.Long.invalidLong(Long.java:124)
12-12 21:08:06.764 10017 10017 E AndroidRuntime: 	at java.lang.Long.parse(Long.java:363)
12-12 21:08:06.764 10017 10017 E AndroidRuntime: 	at java.lang.Long.parseLong(Long.java:353)
12-12 21:08:06.764 10017 10017 E AndroidRuntime: 	at java.lang.Long.parseLong(Long.java:321)
12-12 21:08:06.764 10017 10017 E AndroidRuntime: 	at org.dmfs.tasks.utils.RecentlyUsedLists.getList(RecentlyUsedLists.java:51)
[...]

PR #579, opened 2017-12-13
Fix crash when 'null' had been stored in RecentlyUsedLists. #562

lemonboston:


opentasks/src/main/java/org/dmfs/tasks/utils/RecentlyUsedLists.java
```diff
@@ -39,19 +47,24 @@
...
+            return new EmptyIterable<>();
         }
-        return lists;
+
+        return new Mapped<>(cs -> Long.valueOf(cs.toString()),
+                new Filtered<>(
+                        new Split(strLists.value(), ','),
+                        // 'null' should not be here but happened before: https://github.com/dmfs/opentasks/issues/562
+                        // TODO Checking for null can be removed after a while since stored values will be replaced with correct values on people's devices with this update
+                        Objects::nonNull));
```
dmfs: This check doesn't make any sense. `Split` will not iterate `null`. You need to check for the String `"null"`.
Also the original issue wasn't about a `null` reference, but about `"null"` not being the string representation of a long value.

lemonboston: Oh, that was clearly mistaken. Fixed it.

opentasks/src/main/java/org/dmfs/tasks/utils/RecentlyUsedLists.java
```diff
@@ -105,9 +118,16 @@ public static Long getRecentFromList(Context context, List<Long> allowedLists)
      */
     public static void use(Context context, Long listId)
```
dmfs: we should find out how a `null` could get into the recently used lists in the first place. In think the first step should be to change the `listId` parameter here to `long`. If, at any point, `null` is passed this will result in a `NullPointerException`.

opentasks/src/main/java/org/dmfs/tasks/utils/RecentlyUsedLists.java
```diff
@@ -105,9 +118,16 @@ public static Long getRecentFromList(Context context, List<Long> allowedLists)
...
     {
-        List<Long> lists = getList(context);
-        lists.remove(listId); // does nothing, if "listId" is not in "lists"
-        lists.add(0, listId);
-        setList(context, lists);
+        // 'null' should not come here but happened before: https://github.com/dmfs/opentasks/issues/562
+        if (listId == null)
```
dmfs: This is the opposite of "fail fast" ;-)

opentasks/src/main/java/org/dmfs/tasks/utils/RecentlyUsedLists.java
lemonboston: Do you mean to have that in the next release, so if this happens, we at least may have a stacktrace of the crash?

I spent about 30 minutes now trying to figure out how it can happen without success. Since this `ContentSet` in `EditTaskFragment` can come from multiple places it's hard to track it down. Let me know if I should spend more time with it.

I've also looked into 702994023bf6873cf72865e74098e453c20cdcdc a bit..
(That big change in `TaskListWidgetProvider` doesn't look very nice.)
How about maybe re-implementing this recently used task feature? As I understand it is only a convenience thing for the users, so if we disregard the already stored sharedprefs entry on people's devices and start to collect the data in a new prefs entry again, it wouldn't be a big problem.
Not sure if that would solve the 'null' issue though, I haven't looked into that much in that original changeset.

dmfs: Yes we should do that in the next release. This should be a `long` anyway, there is no point in allowing `null` in the recently used lists.
I don't think we should spend more time on debugging this. Let's make sure it fails fast when something is wrong. That also improves chances of discovering the bug during monkey tests.

lemonboston:
@dmfs I've pushed a new version (rebased) with tests for checking several things regarding this fix.
If you revert the changes in `RecentlyUsedLists` you can see that the tests fail.

Ready for check.

Issue #411, opened 2017-08-29
Add tel URI support


sjulien:
Sometimes, I need to create a "phone call" task.

In the "URL" field, it's impossible to encode a phone number URI : the app says "Invalid URL".

Add the possibility to encode such URIs (see tel URI scheme, RFC 3966 : [https://tools.ietf.org/html/rfc3966](https://tools.ietf.org/html/rfc3966)).

If the user then taps on this URL, the device normally starts a call automatically.

dmfs:
Strictly spoken a URI like `tel:123456789` is not a "URL". The iCalendar property which this field is based on is called [`URL`](https://tools.ietf.org/html/rfc5545#section-3.8.4.6). On the other hand the value of the `URL` field is defined as `URI` which is a much broader term. Following this, I guess it would make sense to allow any valid URI in the `URL` field.

Btw, the app should hyperlink any phone number in the description field, so you might add the number to the description just as well. Not sure if other applications support that though.

PR #580, opened 2017-12-13
Allow URIs in place of the URI field. #411

lemonboston:


opentasks/src/main/res/values-de/strings.xml
```diff
@@ -52,7 +52,7 @@
     <!-- Task editor -->
     <string name="activity_edit_task_title">Bearbeiten</string>
     <string name="activity_add_task_title">Erstellen</string>
-    <string name="activity_edit_task_error_invalid_url">Ungültige URL</string>
+    <string name="activity_edit_task_error_invalid_uri">Ungültige URI</string>
```
dmfs: I don't think we should change the translations. The term "URL" probably makes more sense than "URI" to most users.

lemonboston: Ok, I've reverted those changes, and squashed-rebased so that there is only one commit with the reduced changeset. Ready for check.

Issue #589, opened 2017-12-21
Re-use code of opening task details screen


lemonboston:
This code is repeated at several places:
```java
                Intent viewTaskIntent = new Intent(Intent.ACTION_VIEW);
                viewTaskIntent.setData(mSelectedTaskUri);
                startActivity(viewTaskIntent);

```
Make it like this:
```java
        new BasicTaskDetailsUi(mSelectedTaskUri).show(this);
```

(Note: this comes out the #469 changeset)

PR #591, opened 2017-12-21
Extract and re-use details screen starting code. #589

lemonboston:


opentasks/src/main/java/org/dmfs/tasks/detailsscreen/TaskDetailsUi.java
```diff
@@ -0,0 +1,33 @@
...
+
+import android.content.Context;
+
+
+/**
+ * Represents the task details UI which can be shown.
+ *
+ * @author Gabor Keszthelyi
+ */
+public interface TaskDetailsUi
```
dmfs: The interface looks quite generic. Why not just call it after something that can launch any activity? I mean, how would the same interface for the task editor?

lemonboston: You're right, this can be more generic. How about something like this?:

```java
interface Showable<C extends Context>
{
    void show(C context);
}
```

It could represent a 'meaningful' UI of the app that can be shown. By 'meaningful' I mean that it represents a use-case, like 'details of the task', or like a 'permission request dialog'. Something that makes sense on its own and may be fired/shown independently of the place(s) where it is used.

The followings could be covered by this for example:
- opening a new screen by starting an `Activity` (C as Context)
- showing a `DialogFragment`  (C as Activity to call getFragmentManager())
- showing a `Toast` (C as Context)
- showing a `SnackBar` ([using](https://stackoverflow.com/a/8817003/4247460) findViewById(..content) (C as Activity)
- showing a `Notification` (using activity.getSystemService(NotificationManager).notify(n)) (C as Context)

Fragment transactions would not be included. (Except for `DialogFragment`)

Of course we wouldn't need to actually start to use it everywhere, I mentioned them just for the concept.
Let me know if you think that interface makes sense, I would update the code in this PR in that case.

lemonboston:
I've updated this with using that `Showable` interface, and squashed and rebased.

I also noticed however that there are 5 more places in the codebase where `Intent`s are created with `Intent.ACTION_VIEW`, so those should probably be reviewed as well. But that leads to questions on how we should generally handle these `Intent`s and extras.
So now I am not sure that the changes in this pull request makes sense to merge, it introduces something new in that matter which may be inconsistent with usage elsewhere.

More specifically, we could create `Single<Intent>`s but I am not sure if that's good. Also, the `Box` solution (with `Key`s and `Argument`) could also effect how it is good to approach this.

It probably makes sense to put this on hold, and discuss it before proceeding.

Issue #464, opened 2017-10-18
Allow update and delete operations on the instances table and transalate them to task operations


dmfs:

dmfs:
done

PR #602, opened 2017-12-30
Introduce operations on Instances, implements #464

dmfs:
This commit adds initial support for operations on the instances table. At present sync adapters are not allowed to write to the instances table, non-sync adapters can only insert or modify task values. Instances table columns are still read-only. This is meant to simplify working with recurrence in the UI. The UI won't have to take special care about creating exceptions. It can simply update or delete individual instances and the provider will take care of creating overrides or ex-dates.

### DIFF
--- a/opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
+++ b/opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
@@ -30,13 +30,13 @@ import org.dmfs.android.contentpal.RowSnapshot;
 import org.dmfs.android.contentpal.Table;
 import org.dmfs.android.contentpal.operations.Assert;
 import org.dmfs.android.contentpal.operations.BulkDelete;
+import org.dmfs.android.contentpal.operations.BulkUpdate;
 import org.dmfs.android.contentpal.operations.Delete;
 import org.dmfs.android.contentpal.operations.Put;
 import org.dmfs.android.contentpal.operations.Referring;
-import org.dmfs.android.contentpal.predicates.AllOf;
-import org.dmfs.android.contentpal.predicates.EqArg;
-import org.dmfs.android.contentpal.predicates.IsNull;
+import org.dmfs.android.contentpal.predicates.ReferringTo;
 import org.dmfs.android.contentpal.queues.BasicOperationsQueue;
+import org.dmfs.android.contentpal.rowdata.CharSequenceRowData;
 import org.dmfs.android.contentpal.rowdata.Composite;
 import org.dmfs.android.contentpal.rowdata.EmptyRowData;
 import org.dmfs.android.contentpal.rowsnapshots.VirtualRowSnapshot;
@@ -55,6 +55,7 @@ import org.dmfs.opentaskspal.tasks.StatusData;
 import org.dmfs.opentaskspal.tasks.SyncIdData;
 import org.dmfs.opentaskspal.tasks.TimeData;
 import org.dmfs.opentaskspal.tasks.TitleData;
+import org.dmfs.opentaskstestpal.InstanceTestData;
 import org.dmfs.rfc5545.DateTime;
 import org.dmfs.rfc5545.Duration;
 import org.dmfs.tasks.contract.TaskContract.Instances;
@@ -68,6 +69,7 @@ import org.junit.runner.RunWith;
 import java.util.TimeZone;
 
 import static org.dmfs.android.contenttestpal.ContentMatcher.resultsIn;
+import static org.dmfs.optional.Absent.absent;
 import static org.junit.Assert.assertThat;
 
 
@@ -76,6 +78,7 @@ import static org.junit.Assert.assertThat;
  *
  * @author Yannic Ahrens
  * @author Gabor Keszthelyi
+ * @author Marten Gajda
  */
 @RunWith(AndroidJUnit4.class)
 public class TaskProviderTest
@@ -130,7 +133,7 @@ public class TaskProviderTest
 
 
     /**
-     * Create 1 local task list and 1 task, check values in TaskLists, TaskList, Instances tables.
+     * Create 1 local task list and 1 task, check values in TaskLists, Tasks, Instances tables.
      */
     @Test
     public void testSingleInsert()
@@ -145,17 +148,62 @@ public class TaskProviderTest
         ), resultsIn(mClient,
                 new Assert<>(taskList, new NameData("list1")),
                 new Assert<>(task, new TitleData("task1")),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task, new AllOf(
-                        new IsNull(Instances.INSTANCE_START),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new IsNull(Instances.INSTANCE_DUE),
-                        new IsNull(Instances.INSTANCE_START_SORTING),
-                        new IsNull(Instances.INSTANCE_DUE_SORTING),
-                        new IsNull(Instances.INSTANCE_DURATION),
-                        new IsNull(Tasks.TZ),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, 0)
-                ))
-        ));
+                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new Composite<Instances>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TZ, null))
+                )));
+    }
+
+
+    /**
+     * Create 1 local task list and 1 task (via the instances table), check values in TaskLists, Tasks, Instances tables.
+     */
+    @Test
+    public void testInsertSingleInstance()
+    {
+        RowSnapshot<TaskLists> taskList = new VirtualRowSnapshot<>(new LocalTaskListsTable(mAuthority));
+        RowSnapshot<Instances> instance = new VirtualRowSnapshot<>(new InstanceTable(mAuthority));
+
+        assertThat(new Seq<>(
+                new Put<>(taskList, new NameData("list1")),
+                // insert a new task straight into the instances table
+                new Referring<>(taskList, Tasks.LIST_ID, new Put<>(instance, new CharSequenceRowData<>(Tasks.TITLE, "task1")))
+
+        ), resultsIn(mClient,
+                new Assert<>(taskList, new NameData("list1")),
+                new AssertRelated<>(new InstanceTable(mAuthority), Tasks.LIST_ID, taskList,
+                        new Composite<>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TITLE, "task1"),
+                                new CharSequenceRowData<>(Tasks.TZ, null))
+                )));
+    }
+
+
+    /**
+     * Create 1 local task list and 1 task, update task via instances table and check values in TaskLists, Tasks, Instances tables.
+     */
+    @Test
+    public void testSingleInsertUpdateInstance()
+    {
+        RowSnapshot<TaskLists> taskList = new VirtualRowSnapshot<>(new LocalTaskListsTable(mAuthority));
+        RowSnapshot<Tasks> task = new VirtualRowSnapshot<>(new TaskListScoped(taskList, new TasksTable(mAuthority)));
+        Table<Instances> instancesTable = new InstanceTable(mAuthority);
+
+        assertThat(new Seq<>(
+                new Put<>(taskList, new NameData("list1")),
+                new Put<>(task, new TitleData("task1")),
+                new BulkUpdate<>(instancesTable, new CharSequenceRowData<>(Tasks.TITLE, "task updated"), new ReferringTo<>(Instances.TASK_ID, task))
+
+        ), resultsIn(mClient,
+                new Assert<>(taskList, new NameData("list1")),
+                new Assert<>(task, new TitleData("task updated")),
+                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new Composite<Instances>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TZ, null))
+                )));
     }
 
 
@@ -185,37 +233,22 @@ public class TaskProviderTest
                 new Assert<>(task1, new TitleData("task1")),
                 new Assert<>(task2, new TitleData("task2")),
                 new Assert<>(task3, new TitleData("task3")),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task1, new AllOf(
-                        new IsNull(Instances.INSTANCE_START),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new IsNull(Instances.INSTANCE_DUE),
-                        new IsNull(Instances.INSTANCE_START_SORTING),
-                        new IsNull(Instances.INSTANCE_DUE_SORTING),
-                        new IsNull(Instances.INSTANCE_DURATION),
-                        new IsNull(Tasks.TZ),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, 0)
-                )),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task2, new AllOf(
-                        new IsNull(Instances.INSTANCE_START),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new IsNull(Instances.INSTANCE_DUE),
-                        new IsNull(Instances.INSTANCE_START_SORTING),
-                        new IsNull(Instances.INSTANCE_DUE_SORTING),
-                        new IsNull(Instances.INSTANCE_DURATION),
-                        new IsNull(Tasks.TZ),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, 0)
-                )),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task3, new AllOf(
-                        new IsNull(Instances.INSTANCE_START),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new IsNull(Instances.INSTANCE_DUE),
-                        new IsNull(Instances.INSTANCE_START_SORTING),
-                        new IsNull(Instances.INSTANCE_DUE_SORTING),
-                        new IsNull(Instances.INSTANCE_DURATION),
-                        new IsNull(Tasks.TZ),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, 0)
-                ))
-        ));
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task1,
+                        new Composite<Instances>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TZ, null))),
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task2,
+                        new Composite<Instances>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TZ, null))),
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task3,
+                        new Composite<Instances>(
+                                new InstanceTestData(0),
+                                new CharSequenceRowData<>(Tasks.TZ, null))
+                )));
     }
 
 
@@ -237,17 +270,16 @@ public class TaskProviderTest
 
         ), resultsIn(mClient,
                 new Assert<>(task, new TimeData(start, due)),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task, new AllOf(
-                        new EqArg(Instances.INSTANCE_START, start.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, 0)
-                ))
-        ));
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new Composite<Instances>(
+                                new InstanceTestData(
+                                        start.shiftTimeZone(TimeZone.getDefault()),
+                                        due.shiftTimeZone(TimeZone.getDefault()),
+                                        absent(),
+                                        0),
+                                new CharSequenceRowData<>(Tasks.TZ, "UTC"))
+                )));
     }
 
 
@@ -270,17 +302,16 @@ public class TaskProviderTest
                 new Put<>(task, new StatusData(Tasks.STATUS_COMPLETED))
         ), resultsIn(mClient,
                 new Assert<>(task, new TimeData(start, due)),
-                new AssertRelated<>(new InstanceTable(mAuthority), Instances.TASK_ID, task, new AllOf(
-                        new EqArg(Instances.INSTANCE_START, start.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.DISTANCE_FROM_CURRENT, -1)
-                ))
-        ));
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new Composite<Instances>(
+                                new InstanceTestData(
+                                        start.shiftTimeZone(TimeZone.getDefault()),
+                                        due.shiftTimeZone(TimeZone.getDefault()),
+                                        absent(),
+                                        -1),
+                                new CharSequenceRowData<>(Tasks.TZ, "UTC"))
+                )));
     }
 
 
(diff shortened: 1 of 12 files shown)


dmfs:
@lemonboston note, this goes to the new `recurrence/master` branch, because it's not ready for release until some of the TODOs have been completed and a few more provider tests have been added.
I'll create separate issues for these TODOs if they don't exist yet.

dmfs:
this needs to be rebased on master after #607 has been merged.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -145,16 +148,54 @@ public void testSingleInsert()
...
+                new Put<>(taskList, new NameData("list1")),
+                // insert a new task straight into the instances table
+                new Referring<>(taskList, Tasks.LIST_ID, new Put<>(instance, new CharSequenceRowData<>(Tasks.TITLE, "task1")))
+
+        ), resultsIn(mClient,
+                new Assert<>(taskList, new NameData("list1")),
+                new AssertRelated<>(new InstanceTable(mAuthority), Tasks.LIST_ID, taskList,
+                        new Composite<>(
+                                new InstanceData(Instances.INSTANCE_STATUS_NEXT),
+                                new CharSequenceRowData<>(Tasks.TITLE, "task1"))
```
lemonboston: Shouldn't this test check the Tasks table as well?
Javadoc says "TasksLists, TaskList, Instances" - should that be "TaskLists, Tasks, Instances"?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -145,16 +148,54 @@ public void testSingleInsert()
...
+                )));
+    }
+
+
+    /**
+     * Create 1 local task list and 1 task, update task via instances table and check values in TaskLists, TaskList, Instances tables.
```
lemonboston: Same here with the table list in the javadoc.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -237,16 +251,13 @@ public void testInsertTaskWithStartAndDue()
...
-                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
```
lemonboston: Previous version checked `TZ` as well here.

By the way, there is something basic I don't understand here. Since `Instances` table has all the `TaskColumns` as well, does it mean that all those values are always in sync?
Shouldn't we always test both tables for those values then?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -270,16 +281,13 @@ public void testInsertTaskWithStartAndDueUpdateStatus()
...
-                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_CLOSED)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
```
lemonboston: Same here with the `TZ` check.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -306,16 +314,13 @@ public void testInsertTaskWithStartAndDueMovedForward()
...
-                        new EqArg(Instances.INSTANCE_START_SORTING, startNew.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, dueNew.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, dueNew.getTimestamp() - startNew.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
+                                startNew.shiftTimeZone(TimeZone.getDefault()),
```
lemonboston: Same here with the `TZ` check

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -342,16 +347,13 @@ public void testInsertTaskWithStartAndDueMovedBackwards()
...
```
lemonboston: `TZ`..

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -374,16 +376,13 @@ public void testInsertTaskWithStartAndDueAddedAfterwards()
...
-                        new EqArg(Instances.INSTANCE_DUE, due.getTimestamp()),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
```
lemonboston: `TZ`...

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -407,16 +406,13 @@ public void testInsertWithStartAndDuration()
...
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, start.addDuration(duration).shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new EqArg(Tasks.TZ, "UTC"),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
+                                start.shiftTimeZone(TimeZone.getDefault()),
```
lemonboston: `TZ`..

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -444,16 +440,13 @@ public void testInsertWithStartAndDurationChangeTimeZone()
...
-                        new EqArg(Instances.INSTANCE_DURATION, durationMillis),
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, start.addDuration(duration).shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new EqArg(Tasks.TZ, "America/New_York"),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
```
lemonboston: `TZ`..  here that is the thing under test according to the comment.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -485,16 +478,13 @@ public void testUpdateDue() throws Exception
...
-                        new EqArg(Instances.INSTANCE_START_SORTING, start.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_DUE_SORTING, due2.shiftTimeZone(TimeZone.getDefault()).getInstance()),
-                        new EqArg(Instances.INSTANCE_ORIGINAL_TIME, 0),
-                        new EqArg(Tasks.TZ, "UTC"),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new InstanceData(
+                                start.shiftTimeZone(TimeZone.getDefault()),
```
lemonboston: `TZ`..

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
```diff
@@ -787,6 +798,32 @@ public int deleteInTransaction(final SQLiteDatabase db, Uri uri, String selectio
...
+            case INSTANCE_ID:
+                // add id to selection and fall through
+                selection = updateSelection(selectId(uri), selection);
+
+            case INSTANCES:
+            {
+                // iterate over all instances that match the selection. We iterate "manually" to execute any processors before or after deletion.
+                final Cursor cursor = db.query(Tables.INSTANCE_VIEW, null, selection, selectionArgs, null, null, null, null);
+
+                try
```
lemonboston: Android Studio offers try-with-resources quick fix here now. It's also mentioned in the docs, too, that it is supported:
https://developer.android.com/studio/write/java8-support.html
Shall we start using it?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: That "TaskList" is copy'n'pasted from your method ;-)
We definitely should test the task table values as well, but this is out of scope for this story. This story is really just about testing the instance operations. The `InstanceData` class was required to do that properly (the previous solution with `Predicates` was incorrect and could easily miss actual issues).
We'll certainly add more and better tests before we'll release recurrence.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: No, we actually only need to test the task values in one of both tables. We just have to check the `TASK_ID` in the instances tables. Since we join both tables on this column we can be sure that the task values are correct in the instances view as well (if the `TASK_ID` is correct`). We should just add a test to check if the `join` is correct (i.e. contains all the columns).

In other words we don't have to test that the table join works, we just have to test that it's declared properly.

dmfs:
@lemonboston please check again

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
dmfs: I gave it a try.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/AbstractInstanceAdapter.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * An abstract implementation of a {@link TaskAdapter} to server as the base for more concrete adapters.
+ *
+ * @author Marten Gajda <marten@dmfs.org>
+ */
+public abstract class AbstractInstanceAdapter implements InstanceAdapter
+{
+    @Override
+    public Uri uri(String authority)
```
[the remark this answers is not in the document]
dmfs: fixed

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
```diff
@@ -897,6 +928,17 @@ public Uri insertInTransaction(final SQLiteDatabase db, Uri uri, final ContentVa
 
                 break;
 
+            case INSTANCES:
+            {
+                InstanceAdapter instance = mInstanceProcessorChain.insert(db, new ContentValuesInstanceAdapter(values), isSyncAdapter);
+                rowId = instance.id();
+                result_uri = TaskContract.Tasks.getContentUri(mAuthority);
```
lemonboston: Shouldn't this be `TaskContract.Instances.getContentUri(..)`? Isn't `rowId` the id of the instance, not the task here?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
```diff
@@ -237,17 +270,16 @@ public void testInsertTaskWithStartAndDue()
...
-                        new EqArg(Instances.INSTANCE_DURATION, due.getTimestamp() - start.getTimestamp()),
-                        new EqArg(Tasks.TZ, start.isAllDay() ? "UTC" : start.getTimeZone().getID()),
-                        new EqArg(Instances.INSTANCE_STATUS, Instances.INSTANCE_STATUS_NEXT)
-                ))
-        ));
+                new AssertRelated<>(
+                        new InstanceTable(mAuthority), Instances.TASK_ID, task,
+                        new Composite<Instances>(
+                                new InstanceData(
+                                        start.shiftTimeZone(TimeZone.getDefault()),
```
lemonboston: Why is it needed here to shift the time zone to the default? Aren't we supposed to keep the time zone as it is, which is `UTC` in this case?
The test passes without the shift.
If the shift is needed for the sorting columns, shouldn't that be applied inside `InstanceData`?
(I guess this applies to other test cases as well.)

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/InstanceData.java
```diff
@@ -0,0 +1,85 @@
...
+import org.dmfs.rfc5545.DateTime;
+import org.dmfs.tasks.contract.TaskContract;
+
+import static org.dmfs.optional.Absent.absent;
+
+
+/**
+ * {@link RowData} of the instance table. This sets all values except for {@link TaskContract.Instances#TASK_ID}.
+ * <p>
+ * Note: this is meant for use with an assert operation during tests as the instances table is read only and doesn't allow inserts nor updates.
```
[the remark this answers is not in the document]
lemonboston: Can you imagine using `InstanceData` in production code?
If not and it's only for testing then I think it should be either on test classpath, next to the tests, or in a test module. So what do you think about moving it next to `TaskProviderTest` for now?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
lemonboston: > That "TaskList" is copy'n'pasted from your method ;-)
Oh, I see ;) Would you please correct both then?

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
lemonboston: Ok, I get it now that we use a View for Instances table which joins on Tasks, so they have be the same.
Worth creating a ticket now for testing the join, or not?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // actually, no instance value can be changed, the instance table only allows for updating task values
+        if (instanceAdapter.isUpdated(InstanceAdapter._ID) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_START) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_START_SORTING) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_DUE) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_DUE_SORTING) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_ORIGINAL_TIME) ||
+                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_STATUS) ||
+                instanceAdapter.isUpdated(InstanceAdapter.TASK_ID))
```
lemonboston: As you mentioned on chat, and also as I start to understand `TaskValueDelegate` now, the recurrence fields should be added here, too, to make sure they are not updated. Right?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
```diff
@@ -0,0 +1,167 @@
...
+            task.unset(TaskAdapter.SYNC8);
+            task.unset(TaskAdapter.SYNC_ID);
+            task.unset(TaskAdapter.SYNC_VERSION);
+            // TODO: if this is the first instance (and maybe no other overrides exist), don't create an override but split the series into two tasks
+            mDelegate.insert(db, task, false);
+        }
+        else
+        {
+            // this is a non-recurring task or it's already an override, just delegate the update
+            mDelegate.update(db, taskAdapter, false);
```
lemonboston: What about a direct update to a recurring instance? Wouldn't that lead to here and by simply changing it, braking the data consistency of the recurrence set?
Shouldn't that be detected here if it is a recurring instance?
Or are those going to have `RRULE` or `RDATE` copied from the master, so those are handled by the `if` branch here?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
```diff
@@ -0,0 +1,167 @@
...
+            finally
+            {
+                c.close();
+            }
+        }
+        else if (taskAdapter.isRecurring())
+        {
+            // TODO: if this is the first instance, consider moving the recurrence start instead of adding an exdate
+            addExDate(taskAdapter, entityAdapter.valueOf(InstanceAdapter.INSTANCE_ORIGINAL_TIME));
+            mDelegate.update(db, taskAdapter, false);
```
lemonboston: Is updating `EXDATE` going to trigger a remove of that instance somehow?

Another related question: are the master task instance rows which have the RDATE/RRULE going to be exempt from actual queries, like what to show on the UI? And another instance row will be generated for the starting instance?
I am asking this because otherwise, if the first instance is the master itself, then how can we handle the case when that is removed from the recurrence set? (Which is what this `else if` branch is about if I understand correctly.)

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public InstanceAdapter insert(SQLiteDatabase db, InstanceAdapter entityAdapter, boolean isSyncAdapter)
+    {
+        // for now inserts are just forwarded, but we have to make sure we return an InstanceAdapter which contains the current instance row _ID.
+        // TODO: in case of a recurring master we should add an RDATE to it
+        // TODO: handle when inserting an override that already exists, either update the existing one or fail
+        TaskAdapter taskResult = mDelegate.insert(db, entityAdapter.taskAdapter(), false);
+        Cursor c = db.query(TaskDatabaseHelper.Tables.INSTANCES, new String[] { TaskContract.Instances._ID },
+                TaskContract.Instances.TASK_ID + "=" + taskResult.id(), null, null, null, null);
+        try
```
lemonboston: `try-with-resource` opportunity here (and another one down below).

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: Yes, that's correct.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: Yeah, I guess a ticket would be good, although I'd consider this very low priority, because this is essentially covered by most provider tests.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
dmfs: Not sure what exactly you mean. A recurring instance (which is not an override) will always point to the master task. Which means they are handled by the `if` branch.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
dmfs: Yes, adding an `EXDATE` will result in removing the instance row.

All recurring instances which are not overridden will point to the master task. When you remove the first instance, an `EXDATE` is added and the instance row of the first instance is removed. Nothing else will happen. That reminds me, when all instances are removed we probably should remove the entire task.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
dmfs: done

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
dmfs: good catch, fixed it.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: This is indeed about the sorting columns. IMO `InstanceData` should not convert the time zone. `InstanceData` is just not in the position to decide which time zone to use, there may be cases when you want to test explicitly for a time zone other than the default.

The test passes without shift only if your default timezone is `UTC`. If you change the timezone of your emulator the test should actually fail if remove the shift.

dmfs:
I've rebased this on the current master.

dmfs:
I think this can actually go to `master`. It doesn't really change the current behavior it only adds stuff that's not in use yet.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderTest.java
dmfs: In addition, sometimes the arguments to `InstanceData` need to be calculated, which has to be done in the local time zone (otherwise the result might be wrong), so we can't defer the shift to `InstanceData`.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
lemonboston: Ok, I see now. So I assume 'pointing to the master task' means that the row of them in `Instance_View` are joined on master task's `Task` table row.

So in `Tasks` table we will only have the:
- non-recurring tasks
- recurring masters
- recurring overrides
but no rows for the enumerated recurring instances, right?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
lemonboston: Ok, and regarding the query for the tasks shown in the app, we will exclude the master task row, right? Since we will have a separate instance row for the first instance.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
dmfs: Exactly. The enumerated recurrence instances go to the instances table, which essentially represents the instances as observed by the task UI. Since all non-overridden instances share the same data (except for start and due) it wouldn't make sense to copy the values over and over again.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
dmfs: we don't explicitly exclude the master task. If there happens to be an override for every instance we don't show the master task data, but if some instances have no overrides, they will be joined with the master task (since `TASK_ID` still points there)

opentaskspal/src/test/java/org/dmfs/opentaskspal/tasks/InstanceDataTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertThat(new InstanceData(5),
+                builds(
+                        withValuesOnly(
+                                withNullValue(TaskContract.Instances.INSTANCE_START),
+                                withNullValue(TaskContract.Instances.INSTANCE_START_SORTING),
+                                withNullValue(TaskContract.Instances.INSTANCE_DUE),
+                                withNullValue(TaskContract.Instances.INSTANCE_DUE_SORTING),
+                                withNullValue(TaskContract.Instances.INSTANCE_DURATION),
+                                containing(TaskContract.Instances.INSTANCE_ORIGINAL_TIME, 0L),
+                                containing(TaskContract.Instances.INSTANCE_STATUS, 5)
```
lemonboston: `INSTANCE_STATUS` is not replaced yet here with `DISTANCE_FROM_CURRENT`.

lemonboston: Strange that Travis didn't catch it..

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
lemonboston: Oh, I understand now, so the master task row in Task table doesn't even get a row in Instances table in itself. Only the actual enumerated instances and overrides. Makes sense, considering the 'Instances' name :) (don't know how I could imagine it otherwise)

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/TaskValueDelegate.java
lemonboston: Clear, thanks.

opentaskspal/src/test/java/org/dmfs/opentaskspal/tasks/InstanceDataTest.java
dmfs: apparently it doesn't run these tests.

dmfs: maybe it's because `opentaskspal` doesn't have any Android tests. I've added the `check` task, let's see if that catches the issue.

dmfs: yep, now it's catched. We should check all our Android projects to make sure the `check` task is run.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/InstanceData.java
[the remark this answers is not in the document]
dmfs: Just tried that but I'm facing some problems here. If I move this class to the provider tests, I also have to move the unit test for this class. Unfortunately Robolectric tests don't run as instrumentation tests, I'm getting this error: https://github.com/robolectric/robolectric/issues/3190

For now I'll create a testing package in `opentaskspal` and move the class there.

dmfs:
@lemonboston I've rebased on master and fixed a new more things

opentaskspal/src/test/java/org/dmfs/opentaskspal/tasks/InstanceDataTest.java
lemonboston: I don't understand this, those unit tests should be part of the build.
Docs even say here that 'check' is the default gradle task:
https://docs.travis-ci.com/user/languages/java/
I tried to google it, haven't found any issue mentioned.
Anyway, it's good that it works now.

dmfs: right, `check` is the default, but we have overridden that in the config to run the connected test

lemonboston: oh, I see

PR #604, opened 2017-12-30
Add RowData to create recurring tasks. Implements #460

dmfs:
This certainly needs to be revisited, but it should be ok for now when it's mostly required to write unit tests.

### DIFF
--- a/opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DateTimeListTaskData.java
+++ b/opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DateTimeListTaskData.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2018 dmfs GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dmfs.opentaskspal.tasks;
+
+import android.content.ContentProviderOperation;
+import android.support.annotation.NonNull;
+import android.text.TextUtils;
+
+import org.dmfs.android.contentpal.RowData;
+import org.dmfs.android.contentpal.TransactionContext;
+import org.dmfs.jems.iterable.decorators.Mapped;
+import org.dmfs.rfc5545.DateTime;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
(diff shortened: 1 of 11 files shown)


opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/ExDatesTaskData.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        mExDates = exdates;
+    }
+
+
+    @NonNull
+    @Override
+    public ContentProviderOperation.Builder updatedBuilder(@NonNull TransactionContext transactionContext, @NonNull ContentProviderOperation.Builder builder)
+    {
+        String value = TextUtils.join(",",
```
lemonboston: Why not extract the code duplication between `ExDatesTaskData` and `RDatesTaskData`?

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/ExDatesTaskData.java
```diff
@@ -0,0 +1,56 @@
...
+import org.dmfs.android.contentpal.TransactionContext;
+import org.dmfs.jems.iterable.decorators.Mapped;
+import org.dmfs.rfc5545.DateTime;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * {@link RowData} for tasks with EXDATEs
+ * <p>
+ * TODO: how to make sure this is only ever used with tasks having a start and/or due date?
```
lemonboston: Should that be the responsibility of the provider? To ignore these fields in the input if there is no start/due date.

dmfs: No, a content provider client must send valid data. I want to reduce this kind of fault tolerance to the bare minimum. The provider is supposed to throw if it receives recurrence data for tasks without start or due (at present it doesn't).

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/ExDatesTaskData.java
dmfs: Right, I'll take care of it.

dmfs:
@lemonboston ready for re-review.

opentaskspal/src/main/java/org/dmfs/opentaskspal/tasks/DateTimeListTaskData.java
```diff
@@ -0,0 +1,56 @@
...
+import org.dmfs.rfc5545.DateTime;
+import org.dmfs.tasks.contract.TaskContract;
+
+
+/**
+ * {@link RowData} for a field of {@link DateTime}s.
+ *
+ * @author Marten Gajda
+ */
+public final class DateTimeListTaskData implements RowData<TaskContract.Tasks>
```
lemonboston: Making it package-private perhaps?

dmfs: I chose to generalize it and moved it to its own package. It might be useful in other places (although I don't see any atm). Check again.

Issue #609, opened 2018-01-04
Play Store report: NPE in showFloatingActionButton


dmfs:
```
java.lang.RuntimeException: 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2981)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3042)
  at android.app.ActivityThread.-wrap14 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1639)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6780)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1496)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1386)
Caused by: java.lang.NullPointerException: 
  at org.dmfs.tasks.ViewTaskFragment.showFloatingActionButton (ViewTaskFragment.java:840)
  at org.dmfs.tasks.ViewTaskFragment.loadUri (ViewTaskFragment.java:367)
  at org.dmfs.tasks.TaskListActivity.loadTaskDetailFragment (TaskListActivity.java:462)
  at org.dmfs.tasks.TaskListActivity.onCreate (TaskListActivity.java:242)
  at android.app.Activity.performCreate (Activity.java:6948)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1126)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2924)
```

dmfs:
So far we have reports from Samsung tablets only.

dmfs:
All of them running Android 7.

lemonboston:
Which app version is the stacktrace from? (The lines numbers don't seem to match exactly with current master version.)

dmfs:
Probably the previous release. I'll check if there is another stack trace from the current release.

dmfs:
Here is the stack trace of the latest release:

```
Caused by: java.lang.NullPointerException: 
  at org.dmfs.tasks.ViewTaskFragment.showFloatingActionButton (ViewTaskFragment.java:832)
  at org.dmfs.tasks.ViewTaskFragment.loadUri (ViewTaskFragment.java:365)
[...]

PR #613, opened 2018-01-05
Remove the (duplicate) ViewTaskFragment instantiation upon configurat…

lemonboston:
…ion change. Some cleanup in TaskListActivity. #609 

---

I am submitting this now to allow a preliminary check but I will get back to it and have a second look myself as well.

I couldn't reproduce the bug but the stacktrace suggests that it happens after a configuration change / restoration, when the Activity is started and there is already a `ViewTaskFragment` available instantiated by the framework, and we call `loadUri()` on it, and its views are not initialized yet.
It looks like a clear bug that there is no `savedInstanceState == null` check before doing that.

The log I captured for the rotations with the branch `bugs/609-investigation`:
[Tablet-TaskListActivity-Rotation-Fragments.txt](https://github.com/dmfs/opentasks/files/1607603/Tablet-TaskListActivity-Rotation-Fragments.txt)

I've added that null check in this pull request. I've also removed the `else` branch there which would remove the fragment in case of single pane mode - which I don't see how could happen. @dmfs Please check that part if you can see any case when that makes sense. Can a two pane mode transform into a single pane one?

I've also done some clean up around and removed unused fields, and created a `Unchecked(Single)` which adapts from `Fragile` to avoid the need to catch checked exceptions for a constructor call in `onCreate()`.

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/TaskListActivity.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/TaskListActivity.java
@@ -27,11 +27,9 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.support.annotation.ColorInt;
 import android.support.design.widget.AppBarLayout;
-import android.support.design.widget.CollapsingToolbarLayout;
 import android.support.design.widget.FloatingActionButton;
 import android.support.design.widget.TabLayout;
 import android.support.v4.app.Fragment;
-import android.support.v4.app.FragmentManager;
 import android.support.v4.view.MenuItemCompat;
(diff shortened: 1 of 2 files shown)


dmfs:
Something is not quite working anymore:
Steps to reproduce:
* start a 7" tablet
* create and open a task in portrait mode
* rotate to landscape mode

Expected behavior:
* the same task is opened in the right pane

Actual behavior:
* the task is not opened in the right pane

lemonboston:
This will be handled as part of https://github.com/dmfs/opentasks/issues/624.

Issue #615, opened 2018-01-07
Let provider instances Uri return actual instance `DTSTART` and `DUE` rather than the master task values


dmfs:
Currently, the instances view contains the joined task data as is. When returning the values to the UI it's probably easier if we emulate overrides even for non-overridden recurring instances, which means the task data should contain the actual instance `DTSTART` and `DUE` values instead of the ones of the master task.

dmfs:
Done,

PR #620, opened 2018-01-08
Let instance view return instance DTSTART and DUE rather than the tas…

dmfs:
…k values. Implements #615

This commit improves the instances view by returning the instance start and due times as task `DTSTART` and `DUE`. This way the UI will see the correct instance times.

From the UI perspective the instances view is now a list of individual non-recurring instances, which can be read and written on its own.

Note, for non-recurring tasks, this won't have any effect.


opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
```diff
@@ -1056,14 +1056,15 @@ public static Uri getSearchUri(String authority, String query)
...
-     * {@link Tasks}.
+     * A table containing one entry per task instance. This table is writable in order to allow modification of single instances of a task. Write operations to
+     * this table will be converted into operations on overrides and forwarded to the task table.
      * <p>
-     * TODO: Insert all instances of recurring the tasks.
-     * </p>
+     * Note: The {@link #DTSTART}, {@link #DUE} values of instances of recurring tasks represent the actual instance values, i.e. they are different for each
+     * instance ({@link #DURATION} is always {@code null}).
+     * <p>
+     * Also, none of the instances are recurring themselves, so {@link #RRULE}, {@link #RDATE} and {@link #EXDATE} are always {@code null}.
```
lemonboston: Would it make sense to consider not using all of the `TaskColumns` in `Instances` here?
Since `DURATION`, `RRULE`, `RDATE`, `EXDATE` are always `null`.
Having something like a `CommonTaskColumns`?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
```diff
@@ -166,6 +168,35 @@
...
+            + Tables.TASKS + ".*, "
+            + Tables.LISTS + "." + Tasks.ACCOUNT_NAME + ", "
+            + Tables.LISTS + "." + Tasks.ACCOUNT_TYPE + ", "
+            + Tables.LISTS + "." + Tasks.LIST_OWNER + ", "
+            + Tables.LISTS + "." + Tasks.LIST_NAME + ", "
+            + Tables.LISTS + "." + Tasks.LIST_ACCESS_LEVEL + ", "
+            + Tables.LISTS + "." + Tasks.LIST_COLOR + ", "
+            + Tables.LISTS + "." + Tasks.VISIBLE
+            + " FROM " + Tables.TASKS
+            + " JOIN " + Tables.LISTS + " ON (" + Tables.TASKS + "." + TaskContract.Tasks.LIST_ID + "=" + Tables.LISTS + "." + TaskContract.Tasks._ID + ")"
```
lemonboston: At the end of the first join, shouldn't `TaskContract.Tasks._ID` be `TaskContract.TaskLists._ID` instead? I know it's effectively the same, but just for clarity.

opentaskspal/src/main/java/org/dmfs/opentaskstestpal/InstanceTestData.java
```diff
@@ -79,7 +80,14 @@ public InstanceTestData(@NonNull Optional<DateTime> instanceStart, @NonNull Opti
                 .withValue(TaskContract.Instances.INSTANCE_DURATION,
```
lemonboston: Is the `Instances.INSTANCE_DURATION` column need at all if it is always `DUE` which is used?

dmfs: `Instances.INSTANCE_DURATION` was added for two reasons:

* provide an easy way to the UI to get the actual absolute duration of an instance
* provide an easy way to filter instances by their duration (e.g. in an SQL query)

both could certainly achieved with `Instances.INSTANCE_START` and `Instances.INSTANCE_DUE`, but considering that these might also be `null` (which would be returned as `0` by `Cursor.getLong(int)`), it would require more work work to achieve that without a duration column.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
dmfs: Right, I'll fix that.

opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
dmfs: no, the instances table should be compatible with the tasks table, i.e. instances are special case of a task. In particular, a projection that works on the task table should work on the instances table as well.

dmfs:
@lemonboston I've fixed the _ID column and also mapped the ORIGINAL_INSTANCE_TIME value to the actual instance value. This is a bit more complicated because the instances table contains `0` instead of `null` for absent values.

lemonboston:
And what is the reason for Instances table using `0` instead of `null` for ORIGINAL_INSTANCES_TIME?

opentaskspal/src/main/java/org/dmfs/opentaskstestpal/InstanceTestData.java
lemonboston: Oh, I realized that mistook the override in the VIEW, I though it was INSTANCE_DURATION that is nulled but that's Tasks.DURATION...  Ok, I see it then why this is needed.

opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
lemonboston: ok

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
```diff
@@ -166,6 +168,40 @@
...
+     * ORIGINAL_INSTANCE_TIME values with respective values of the instance.
+     * <p>
+     * This is the instances view as seen by the content provider clients.
+     */
+    private final static String SQL_CREATE_INSTANCE_CLIENT_VIEW = "CREATE VIEW " + Tables.INSTANCE_CLIENT_VIEW + " AS SELECT "
+            + Tables.INSTANCES + ".*, "
+            // override task due and start times with the instance values
+            + Tables.INSTANCES + "." + TaskContract.Instances.INSTANCE_START + " as " + Tasks.DTSTART + ", "
+            + Tables.INSTANCES + "." + TaskContract.Instances.INSTANCE_DUE + " as " + Tasks.DUE + ", "
+            // also replace ORIGINAL_INSTANCE_TIME.
```
lemonboston: Why is it needed to replace ORIGINAL_INSTANCE_TIME? Doesn't it break the specification for that column on the contract? Its value will be different from what the docs says there, if I am not mistaken.

dmfs: No not really. `ORIGINAL_INSTANCE_TIME` is the time at which an instance occurs as per the specification in the master. The instances view represents a flattened view on the tasks, where each task instance is represented as a non-recurring override. Or in other words, it "emulates" an override for each instance of every (recurring or non-recurring) task.
So to me it makes sense to also "emulate" `ORIGINAL_INSTANCE_TIME` if present.

dmfs:
The instances table uses `0` because that's easier to compare, no need to deal with `null` in Java code.

dmfs:
Hmm... there seems to be only one place where it actually matters, I guess I can drop this requirement.

dmfs:
@lemonboston I dropped the requirement of using `0` for `INSTANCE_ORIGINAL_TIME` of non-recurring tasks and simplified that view column.

dmfs:
@lemonboston I've also added some JavaDoc to the Instances contract to explain how the operations will affect the tasks. The insert behavior is not fully implemented yet, at least not in this branch. I'll post a separate PR to fix this.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
lemonboston: This is still a bit confusing for me. Could you please remind me why the `Instances.INSTANCE_ORIGINAL_TIME` column is needed?

The docs for the two columns talk about almost the same things but with a bit different wording. Could it be maybe updated to be phrased the same? It could be easier to see the difference.

lemonboston:
Those javadocs for the Instances table look very useful.

I've checked the changes, no more questions from me besides the one about the ORIGINAL_TIMEs.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskDatabaseHelper.java
dmfs: I'm using `Instances.INSTANCE_ORIGINAL_TIME` primarily as a sorting key for all instanced of a recurring task. Without it I can't sort regular instances and overrides in the correct order. Correct sorting is very helpful when syncing the instances specified in the master with the instances in the database.
Also, with this column it is easier to handle recurring tasks without a start. In this case we assume `start == due` and store that as the `INSTANCE_ORIGINAL_TIME`. Without it, we would have to check at usage time which value we need to use, start or due (because we still store `null` in the start column).

PR #625, opened 2018-01-09
Fix instance insert handling. Implements #623

dmfs:
This commit fixes inserting instances to comply with the JavaDoc. In order to achive that this improves DateTimeArrayFieldAdapter by returning and taking Iterables instead of arrays.


dmfs:
@lemonboston I've come to the conclusion that we don't need a separate `recurrence/master` branch if we implement the stories in the right order. After this has been merged we can essentially switch the UI to using instance URIs and release this. If we did everything right, the user experience should not change at all.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
     public InstanceAdapter insert(SQLiteDatabase db, InstanceAdapter entityAdapter, boolean isSyncAdapter)
     {
         validate(entityAdapter, isSyncAdapter);
-        return mDelegate.insert(db, entityAdapter, isSyncAdapter);
+
+        Optional<Long> instanceId = new NullSafe<>(entityAdapter.taskAdapter().valueOf(TaskAdapter.ORIGINAL_INSTANCE_ID));
+        if (instanceId.isPresent())
+        {
+            String timeStampString = Long.toString(entityAdapter.taskAdapter().valueOf(TaskAdapter.ORIGINAL_INSTANCE_TIME).getTimestamp());
```
lemonboston: Wouldn't it be better to read ORIGINAL_INSTANCE_TIME here as `Optional` as well and throw explicitly in case of absent?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
+            // there better be no instance for this yet, otherwise this is going to fail
```
lemonboston: I am not sure I understand the "better be" phrasing here, it sounds like hoping for, not sure. 
Could this be stated as well?: "If there is an instance already for this id and time, this is going to throw."

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
     public InstanceAdapter insert(SQLiteDatabase db, InstanceAdapter entityAdapter, boolean isSyncAdapter)
```
lemonboston: Could this be refactored to something like this?
```java
    @Override
    public InstanceAdapter insert(SQLiteDatabase db, InstanceAdapter entityAdapter, boolean isSyncAdapter)
    {
        validateNotSyncadapter(isSyncAdapter);
        validateOnlyAllowedFieldsAreUpdated(entityAdapter);
        validateInstanceDoesNotExistAlready(db, entityAdapter);

        return mDelegate.insert(db, entityAdapter, false);
    }
```
The query is only serves as a validation as well, right?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
lemonboston: Wouldn't it be better to read ORIGINAL_INSTANCE_TIME here as Optional as well and throw explicitly in case of absent?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
lemonboston: Could this method be refactored to something like this?:
```java
    @Override
    public InstanceAdapter insert(SQLiteDatabase db, InstanceAdapter entityAdapter, boolean isSyncAdapter)
    {
        validateNotSyncadapter(isSyncAdapter);
        validateOnlyAllowedFieldsAreUpdated(entityAdapter);
        validateInstanceDoesNotExistAlready(db, entityAdapter);
        
        return mDelegate.insert(db, entityAdapter, false);
    }
```
I think it could be clearer and more self-documenting.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
+                            timeStampString,
+                            timeStampString,
+                            timeStampString },
+                    null,
+                    null,
+                    null))
+            {
+                if (c.getCount() > 0)
+                {
+                    throw new UnsupportedOperationException(String.format(Locale.ENGLISH, "Instance %s of task %d already exists",
```
lemonboston: Wouldn't `IllegalArgumentException` or `IllegalStateException` be more appropriate?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
+        if (instanceId.isPresent())
+        {
+            String timeStampString = Long.toString(entityAdapter.taskAdapter().valueOf(TaskAdapter.ORIGINAL_INSTANCE_TIME).getTimestamp());
+            // there better be no instance for this yet, otherwise this is going to fail
+            try (Cursor c = db.query(
+                    TaskDatabaseHelper.Tables.INSTANCE_VIEW,
+                    new String[] { TaskContract.Instances._ID },
+                    // find any instance which refers to the given original ID and has the same instance time
+                    // for recurring tasks this matches the INSTANCE_ORIGINAL_TIME, for non-recurring tasks this matches start or due (whichever is present).
+                    String.format("(%1$s == ? or %2$s == ?) and (%3$s == ? or %3$s is null and %4$s == ? or %3$s is null and %4$s is null and %5$s == ?) ",
```
lemonboston: This query is a bit difficult to follow for me, would you consider splitting it to two queries (recurring, non-recurring)?
Maybe not effectively run two queries, just split the sql query string into two - if that can be done - like this:
```java
// check for existing non-recurring task:
String.format("........", TaskContract.Instances.TASK_ID, TaskContract.Instances.INSTANCE_START, TaskContract.Instances.INSTANCE_DUE)
+ " or " + 
// check for existing recurring instance:
String.format("....", TaskContract.Instances.ORIGINAL_INSTANCE_ID, TaskContract.Instances.INSTANCE_ORIGINAL_TIME,
```

By the way, for me it's not clear whether `String.format()` or string concatenation is easier to read for this sql statements. (Kotlin's string template would be of course be better.)

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: I've been thinking the same, but I wasn't sure. I'll change it.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -43,50 +69,79 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
         // actually, no instance value can be changed, the instance table only allows for updating task values
-        if (instanceAdapter.isUpdated(InstanceAdapter._ID) ||
-                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_START) ||
-                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_START_SORTING) ||
-                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_DUE) ||
-                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_DUE_SORTING) ||
-                instanceAdapter.isUpdated(InstanceAdapter.INSTANCE_ORIGINAL_TIME) ||
-                instanceAdapter.isUpdated(InstanceAdapter.DISTANCE_FROM_CURRENT) ||
-                instanceAdapter.isUpdated(InstanceAdapter.TASK_ID))
+        if (new First<>(new Sieved<>(instanceAdapter::isUpdated, INSTANCE_FIELD_ADAPTERS)).isPresent())
```
lemonboston: Nice with this Iterable-Optional usage for the condition.

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderInstancesTest.java
```diff
@@ -0,0 +1,346 @@
...
+                                new CharSequenceRowData<>(Tasks.ORIGINAL_INSTANCE_TIME, dtstart),
+                                new CharSequenceRowData<>(Tasks.TZ, "UTC"),
+                                new CharSequenceRowData<>(Tasks.TITLE, "task1")))
+
+        )).commit(mClient);
+    }
+
+
+    /**
+     * Create a single instance and insert an override for a new instance, turning the event into a recurring event.
```
lemonboston: What is the use-case for this - turning a single task into a recurrent one by inserting an override?
Wouldn't we always create recurring tasks by creating a master first? Or turning an single task into a master by updating it with rules.
And only then could we create an override by selecting one of the recurring instances and updating it, as I imagine.

I am talking about usage from user point of view, in a calendar app or in this task app.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
```diff
@@ -0,0 +1,207 @@
...
+        if (value != null)
+        {
+            String stringValue = new Reduced<>(
+                    new StringBuilder(256),
+                    StringBuilder::append,
+                    new Alternating<>(
+                            new Mapped<>(
+                                    DateTime::toString,
+                                    new Mapped<>(dt -> dt.isFloating() ? dt : dt.shiftTimeZone(DateTime.UTC), value)),
+                            new ConstantGenerator<>(","))).value().toString();
```
lemonboston: I would love if we could switch to Kotlin in the near future.. This part would become:
```kotlin
fun setIn(values : ContentValues, dateTimes: Iterable<DateTime>) {

    val datesString = dateTimes
            .map { if (it.isFloating) it else it.shiftTimeZone(DateTime.UTC) }
            .joinToString(",")
    ...
}
```

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
```diff
@@ -0,0 +1,207 @@
...
+            values.put(mDateTimeListFieldName, stringValue.isEmpty() ? null : stringValue);
```
lemonboston: How can `stringValue.isEmpty()` be true here?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private final class DateTimeList extends DelegatingIterable<DateTime>
+    {
+
+        public DateTimeList(TimeZone timeZone, String dateTimeList)
+        {
+            super(new Mapped<>(
+                    datetime -> !datetime.isFloating() && timeZone != null ? datetime.shiftTimeZone(timeZone) : datetime,
+                    new Mapped<CharSequence, DateTime>(
+                            charSequence -> DateTime.parse(timeZone, charSequence.toString()),
```
lemonboston: Why is it needed to shift the timezone, doesn't `DateTime.parse()` set it up correctly?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Constructor for a new {@link DateTimeIterableFieldAdapter}.
+     *
+     * @param datetimeListFieldName
+     *         The name of the field that holds the {@link DateTime} list.
+     * @param timezoneFieldName
+     *         The name of the field that holds the time zone name.
+     */
+    public DateTimeIterableFieldAdapter(String datetimeListFieldName, String timezoneFieldName)
```
lemonboston: Could we start using `@NonNull` and `@Nullable` everywhere? The project depends heavily currently on values being `null`, so it could make it clearer. IDE can also show warnings. Also if we happen to migrate to Kotlin, those annotations are very helpful for the interoperatilbity.

In this class specifically, one field is effectively nullable, the other is not, so it would help understand the null checks below.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
dmfs: If the `Iterable` has no elements.

lemonboston: oh, right

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
dmfs: If the String ends with a `Z` the `parse` method will always return a `UTC` `DateTime`, no matter which time zone was passed.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
dmfs: After committing this I noticed that I can just use `TextUtils.join`. I'll change that.

dmfs: Not it's just 
```
String stringValue = TextUtils.join(",", new Mapped<>(dt -> dt.isFloating() ? dt : dt.shiftTimeZone(DateTime.UTC), value));
```

opentasks-provider/src/androidTest/java/org/dmfs/provider/tasks/TaskProviderInstancesTest.java
dmfs: This is not about `turning a single task into a recurrent one` it's about allowing to add instances to any task, (already) recurring or not. From the provider client's point of view the instances table is one big space of instances. It doesn't really matter which one is of a recurring task and which one isn't. And it should allow the user to add an instance to any task (which is not an override already). I'm currently considering to add another column to the instances table which contains the master task ID, so you always have direct access. This would make it easier to handle these cases. My goal is to allow the UI to handle recurring and non-recurring tasks mostly the same way (i.e. "non-recurring" as a special case of "recurring").

As for the use case, think about opening a task and having an option to repeat this task once again, i.e. sort of a "manual recurrence".

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: changed it

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: updated

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: updated

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: neither way appears to be more readable to me, let's keep it like that until we can replace it with something more sophisticated (i.e. Anko or "SQLitePal" ;-) ). As a first step we could port the ContentPal predicates. That shouldn't be to difficult but can help a lot to write meaningful select statements.

dmfs:
@lemonboston please check my changes

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
dmfs: Yeah, ideally we get rid of all the `null` stuff. This class in particular could be split into two, one which reads and writes the timestamps and a decorator which handles the time zones.

Nevertheless, using the annotations is still a good idea.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -42,51 +75,110 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
+            throw new IllegalArgumentException("Recurrence values can not be modified through the instances table.");
+        }
+    }
+
+
+    private void validateOriginalInstanceValues(InstanceAdapter instanceAdapter)
+    {
+        TaskAdapter taskAdapter = instanceAdapter.taskAdapter();
+        // Updates of ORIGINAL_INSTANCE_* fields are not allowed
+        if (new First<>(new Sieved<>(taskAdapter::isUpdated, ORIGINAL_INSTANCE_FIELD_ADAPTERS)).isPresent())
```
lemonboston: Copy-paste in the exception message below.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
```diff
@@ -42,51 +75,110 @@ public Validating(EntityProcessor<InstanceAdapter> delegate)
...
-                taskAdapter.isUpdated(TaskAdapter.RDATE) ||
-                taskAdapter.isUpdated(TaskAdapter.EXDATE))
+        if (new First<>(new Sieved<>(taskAdapter::isUpdated, RECURRENCE_FIELD_ADAPTERS)).isPresent())
+        {
+            throw new IllegalArgumentException("Recurrence values can not be modified through the instances table.");
+        }
+    }
+
+
+    private void validateOriginalInstanceValues(InstanceAdapter instanceAdapter)
```
lemonboston: Is there a test case for this already?
I suppose all fields would need to be tested individually, for insert and update, so that may need some re-usable solution. And I also know that more tests are probably coming in the coming PRs, but I am a bit concerned that it's easy to loose track what to get back to. So perhaps start adding TODOs for test-cases in the test class?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
lemonboston: It's nice, I think

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
lemonboston: That's good, minus 3 classes :)

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
lemonboston: There is only one test for the non-recurring case currently, right? Can we cover the other case(s) as well, or add TODOs in the test?

lemonboston: I can see that select string builder with the predicates would be useful here.
I've just checked and Anko SQLite doesn't seem to have anything for that.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
lemonboston: I see. It's seems a bit difficult and thus error prone to keep track of the proper ways of `DateTime` conversions.
How about creating more elemental `Single`s for them and collecting them in a dedicated package?
The `Single<DateTime>` which could be extracted from here would take a `TimeZone` and a string timestamp.
For me that would be a much clearer situation, to have that few classes defined somewhere which are for sure right, and can be re-used. The goal would be to not have to shift, swap timezone and things like that anywhere else in code.

This could be a separate ticket of course. There is already some work in progress related to this on my side as well.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: The recurrence tests are part of another story (the one which implements the recurrence expansion). I already implemented most of them.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/model/adapters/DateTimeIterableFieldAdapter.java
dmfs: Right, we can certainly do that. I was also planning on reworking the DateTime class into something more SRP-ish.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: done #626

opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/instances/Validating.java
dmfs: fixed

Issue #467, opened 2017-10-24
Deprecation of android.text.format.Time


frankFic:
Hi,
I found that class android.text.format.Time is used in many places in open tasks (like [NotificationActionUtils](https://github.com/dmfs/opentasks/blob/6d5db7bc045e4a74cedbea4461719637714a4897/opentasks/src/main/java/org/dmfs/tasks/notification/NotificationActionUtils.java#398), [NotificationUpdaterService](https://github.com/dmfs/opentasks/blob/6d5db7bc045e4a74cedbea4461719637714a4897/opentasks/src/main/java/org/dmfs/tasks/notification/NotificationUpdaterService.java#L588), [TimeRangeCursorFactory](https://github.com/dmfs/opentasks/blob/7eb211a910cd607bb04e93ab9270813c38e9bc9a/opentasks/src/main/java/org/dmfs/tasks/groupings/cursorloaders/TimeRangeCursorFactory.java#L117) and many other places). This class has been deprecated since API since API 22 due to many known issues, including 2038 problem, timezone problem etc. (details in the [doc](https://developer.android.com/reference/android/text/format/Time.html)) As suggested by the doc, [GregorianCalendar](https://developer.android.com/reference/java/util/GregorianCalendar.html) will be a good alternative for Time.

dmfs:
Right, we already started to get rid of this. The provider doesn't use `Time` anymore. We've replaced it with our own `DateTime` implementation and we intend to do the same in the UI module. We'll probably do this when we start to integrate our new date-time picker.

PR #646, opened 2018-01-23
Remove android.text.format.Time usage. #467

lemonboston:
@dmfs I am submitting this just to start discussing the changes, what to extract, etc. I haven't reviewed the changeset yet.

Notes, questions, comments:

1. New classes, types to for jems: `Conditional`, `MapFlattened` (flatmap), `Procedure`
I am going to add tickets and pull requests in jems, we can discuss the names, details there.

2. Unused classes that used `Time`: `NotAfter`, `NotBefore`, `ShiftIfAfter`, `ShiftTime` - okay to remove them?

.. there are many more questions, general comments, I will put them down later.
I also plan to change some of the new classes, for example to introduce `Evaluated` decorator where `Cursor` is passed in. Also not use that new `ContainerArgument` but the `Single<>` decoration as you did in the provider.


lemonboston:
The changed `TimeFieldEditor` can produce `DateTime`s with `null` timezones currently - not sure if that's correct or not, I am yet to review that. But this is a general question as well, since `Time` couldn't represent floating times 'UTC' or default time zone was used everywhere. When I did the changes, I tried to keep everything the same (apparently I couldn't at this stage with `TimeFieldEditor`).

A concrete issue now is that now there is a crash with the current state of `TimeFieldEditor` because of this check:
https://github.com/dmfs/opentasks/blob/aa955cd017a9e5efce0ca923616a111deffd68c4/opentasks-provider/src/main/java/org/dmfs/provider/tasks/processors/tasks/Validating.java#L265-L270

Is this check still relevant or shall we remove it start to allow floating times? (Besides reviewing `TimeFieldEditor`.)

lemonboston:
There are several new `Single<DateTime>` classes created here, some of which are totally general, not related to the provider. Like these:
```
StartOfHour
StartOfNextDay
StartOfNextHour
StartOfNextMonth
StartOfNextYear
MovedToDate
MovedToTimeOfDay
MillisDuration
```
They are yet to be reviewed together with their place of usage, so they may change.
I'd like to suggest that we create a small `datetime` module in the project for them. This would also be the only module that imports `org.dmfs:rfc5545-datetime`.

As far as I know gradle builds multi-module projects faster now compared to having the same code in single-module because of incremental builds (doesn't recompile untouched modules). So from this point of view it is beneficial to modularize the project more.

What do you think, are you okay with creating that datetime module? Once these classes seemed to be settled here, meaning we are sure they are needed, this could be taken into a separate ticket, where tests can be added, too.

lemonboston:
Note that as per current state the new jems types are not needed, only `Procedure` but that doesn't require a jems release necessarily.

lemonboston:
In order to break this up to smaller pull requests meaningfully, I am going to create a ticket for the new DateTimeFieldAdapters (Combined and Simple). I will continue on a new branch with the stripped down changeset for that pull request and put this branch on hold.

Issue #644, opened 2018-01-22
Add NonNull & Nullable annotations to FieldAdapter method parameters


lemonboston:
It's about this class: `org.dmfs.tasks.model.adapters.FieldAdapter`

---

It is not obvious for the `getDefault()` method's `ContentSet` parameter which one it should be by looking at the usage. `TimezoneFieldAdapter` passes in `null`:
https://github.com/dmfs/opentasks/blob/7eb211a910cd607bb04e93ab9270813c38e9bc9a/opentasks/src/main/java/org/dmfs/tasks/model/adapters/TimezoneFieldAdapter.java#L101-L106

But otherwise, I would have assumed it should be `@NonNull`, the same as for the other methods. Which one should it be?

PR #649, opened 2018-01-24
Update FieldAdapter with Nullable/NonNull annotations. Fix code in im…

lemonboston:
…plementations where it was needed. #644

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/model/adapters/BooleanFieldAdapter.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/model/adapters/BooleanFieldAdapter.java
@@ -18,9 +18,12 @@ package org.dmfs.tasks.model.adapters;
 
 import android.content.ContentValues;
 import android.database.Cursor;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import org.dmfs.tasks.model.ContentSet;
 import org.dmfs.tasks.model.OnContentChangeListener;
+import org.dmfs.tasks.utils.BooleanBinaryLong;
 
 
 /**
@@ -79,7 +82,7 @@ public final class BooleanFieldAdapter extends FieldAdapter<Boolean>
 
 
     @Override
-    public Boolean get(ContentSet values)
+    public Boolean get(@NonNull ContentSet values)
     {
         Integer value = values.getAsInteger(mFieldName);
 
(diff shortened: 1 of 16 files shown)


opentasks/src/main/java/org/dmfs/tasks/model/adapters/BooleanFieldAdapter.java
```diff
@@ -100,35 +103,35 @@ public Boolean get(Cursor cursor)
...
         return mDefaultValue;
     }
 
 
     @Override
-    public void set(ContentSet values, Boolean value)
+    public void set(@NonNull ContentSet values, @Nullable Boolean value)
     {
-        values.put(mFieldName, value ? 1 : 0);
+        values.put(mFieldName, new BooleanBinaryLong(value).value());
```
dmfs: Why long?

lemonboston: The reason why I started to use `Long` instead of `String` for 0/1 booleans is that I saw that in SQLite these booleans are actually stored as number. I just looked up again:
https://www.sqlite.org/datatype3.html#boolean_datatype
I see integer can be 8 bytes there, so java long would cover it, but for this 0/1, java integer could fit as well, I suppose.

This also relates to the type I've introduce in the other pull request:
https://github.com/dmfs/opentasks/blob/e35047c9791b6ad0b035be17b4961cb5f87e7af4/opentaskspal/src/main/java/org/dmfs/opentaskspal/datetimefields/DateTimeFields.java#L31-L54
So the all-day flag is represented as `Long` here as well, as its raw value.

So in order to keep this consistent, I chose `Long` here, too. Although, I remember that some container types only store everything in `String` (Which type was that? I forgot. ) which may mean a double conversion. So it's not that obvious what's best.
But the idea again was to use the raw type and keep it consistent for simplicity.

What do you think?
We could certainly change to `Integer` at least.

lemonboston: Or we could even use `Short`. (I've never used that type so far..:) )

lemonboston: `Short` is not supported by all containers, so that's not good. `Integer` is supported.

lemonboston: I've changed this to use `Integer`.

PR #651, opened 2018-01-25
Add nullability annotations to ContentSet methods. #650

lemonboston:


opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
```diff
@@ -336,13 +340,13 @@ private ContentValues ensureAfter()
             values = new ContentValues();
             mAfterContentValues = values;
             // also create mAfterKeys
-            mAfterKeys = new HashSet<String>();
+            mAfterKeys = new HashSet<>();
```
dmfs: Please remove `mAfterKeys` while you're at it. It was only a workaround for the absence of `ContentValues.keySet()` on Android 2.x. Since we don't support Android 2 any more you can remove it in most places and replace by `mAfterContentValues.keySet()` where it's being read.

opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
```diff
@@ -544,7 +553,7 @@ else if (mBeforeContentValues != null && mBeforeContentValues.get(key) != null)
     }
 
 
-    public void addOnChangeListener(OnContentChangeListener listener, String key, boolean notify)
+    public void addOnChangeListener(@NonNull OnContentChangeListener listener, @Nullable String key, boolean notify)
```
dmfs: I think it doesn't make sense to add a listener for changes to a `null` key. So `key` should be `@NonNull`.

opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
```diff
@@ -563,7 +572,7 @@ public void addOnChangeListener(OnContentChangeListener listener, String key, bo
     }
 
 
-    public void removeOnChangeListener(OnContentChangeListener listener, String key)
+    public void removeOnChangeListener(@NonNull OnContentChangeListener listener, @Nullable String key)
```
dmfs: same here, `key` is `@NonNull`.

opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
lemonboston: There are usages of these methods with `null` key, so that's why I used `Nullable`. I've checked now and `null` actually has a special usage here, see:
https://github.com/dmfs/opentasks/blob/4453861aa3882ae2ed8e6ca48354adfcfdb56563/opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java#L69-L73

So now I've just added some javadoc to the methods about the key parameter.
Let me know if you'd prefer it to be changed somehow.

opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
lemonboston: Ok, I've removed `mAfterKeys`.

opentasks/src/main/java/org/dmfs/tasks/model/ContentSet.java
dmfs: I right, I forgot about that. Passing null will cause the listener to be called for any update on any key. Did you check if we actually use that?

lemonboston: Yes, they are called with `null` key from `ViewTaskFragment` and `EditTaskFragment`.

PR #679, opened 2018-04-26
Updated Dutch translation by Toon van Gerwen

paulgevers:


opentasks/src/main/res/values-nl/strings.xml
```diff
@@ -64,32 +64,32 @@
     <string name="task_start">Start</string>
     <string name="task_due">Einde</string>
     <string name="task_completed">Afgerond</string>
-    <string name="task_percent_complete">Percentage af</string>
+    <string name="task_percent_complete">Voortgang</string>
     <string name="task_status">Status</string>
-    <string name="task_classification">Privacy</string>
+    <string name="task_classification">Samenwerking</string>
```
dmfs: Sorry for the delay. I've been quite busy during the last couple of months. I've just reviewed these changes and I agree with most of them. This is the only one I don't really like. `Samenwerking` doesn't sound quite right because it has this connotation of collaboration or working jointly on a topic. The meaning of this setting, however, is about privacy.
So what do you think about "Zichtbaarheid"? I think this is a common term in this context. At least in German "Sichtbarkeit" is a reasonable choice and I'm inclined to update the German translation to "Sichtbarkeit".

dmfs:
This is now covered by #710

PR #683, opened 2018-05-23
Add QuickSettingsTile for easy task creation

bephinix:
This MR adds a simple QuickSettingsTile for OpenTasks, so you are able to create a task from your quick settings on the fly.


wevcode:
@dmfs What do you think about it?

dmfs:
I don't have any experience with these quick setting tiles on Android N yet. The idea sounds great. I'll check it out.

wevcode:
@dmfs It is basically only a shortcut to a new task.

thomas-mc-work:
Would it then be possible to add this as a widget on the home screen? That'd be great!

PR #699, opened 2018-09-13
automated code inspect

BO41:
Automated code inspection by android studio. Fixing problems AS can fix automatically

part 2


BO41:
@dmfs okay, thanks.
I don't want to make huge PRs because they are hard to review. But this time it would be better because otherwise many conflicts occur.

I can make a new one, no problem. Either once you reviewed the others or after some other things got merged :)

PR #717, opened 2018-11-02
Added trigger reason due to restart

christianlupus:


documentation/concepts/notifications/notification-concept.adoc
```diff
@@ -16,6 +16,7 @@ This concept describes how notifications should be handled by OpenTasks.
 * Notification texts (ticker text, title, description ...)
 * Notification actions
 * Notification layout (use inbox style to show task text?)
+* Saving of task state during reboot (open reminder, deferred, confirmed)
```
dmfs: Right, that's absolutely something we have to consider.

documentation/concepts/notifications/notification-concept.adoc
```diff
@@ -57,6 +58,11 @@ A task has been deferred::
 
 The task has been deferred, either in the UI, a notification or "remotely" by a sync adapter.
 
+The device boots up::
```
dmfs: The intent is good, but this is the wrong place IMO. This section is about task properties which may result in a notification, i.e. for a notification to show up on boot it still requires one of these triggers having "occurred" either before shut down or during off-time. The boot event alone is not sufficient for a notification be created.

Restoring notifications on boot deserves its own section. Also, we should not just show a notification for all tasks which have started in the past. Some of these may have been acknowledged before the device was shut down. It could be annoying if they show up again on every reboot. So we need to maintain a state of visible notifications (hence the state diagram).

christianlupus: I am sorry, then I understood this wrongly. I thought you would collect all events that might case an notification to be fired. Of course a check needs to be carried out in order to avoid reissuing the same events over and over.
This is also the reason for the first change in the file.

dmfs:
@christianlupus please check the current version of the notification concept branch. I've made a couple of updates which should cover this PR as well.

christianlupus:
@dmfs I read through the document and added some hints where I see a minor enhancement. I hope I made clear what I was thinking (see commit message).

documentation/concepts/notifications/notification-concept.adoc
```diff
@@ -127,30 +127,32 @@ A task provider operation::
...
 
-   After booting the system no notifications are shown and wwe have to restore them.
+   After booting the system no notifications are shown and we have to restore them.
    Ideally we also restore start and due notifications which have not been
    dismissed yet.
 
 The app has been updated::
 
-   When the app is updated, it is first stopped, causing exiting notifications to be
+   When the app is updated, it is first stopped, causing existing notifications to be
```
dmfs: All OpenTasks notifications are exiting, aren't they? ;-)

documentation/concepts/notifications/notification-concept.adoc
```diff
@@ -127,30 +127,32 @@ A task provider operation::
 
     We don't distinguish between UI operations and sync adapter operations. We should
     check if a "notification relevant" field has been updated. This way we can avoid
-    notification updates in case only a field like "title" has been changed.
+    notification updates in case only a field like "description" has been changed.
```
dmfs: Oh, true. Notifications show the title :-o

documentation/concepts/notifications/notification-concept.adoc
christianlupus: Yes ;-)

documentation/concepts/notifications/notification-concept.adoc
christianlupus: This is one "bug" I am suffereing right now but that is of minor importance as it happens only rarely.

christianlupus:
These are my thoughts. The rest seems good to go for me. I did not revisit the state diagram, if this was needed. There might be some modifications needed, but here I am not very experienced about the code to help much, sorry.

dmfs:
Thanks. I still need a concept for the notification content and layout, i.e. what to show, how should it be arranged and look like. The challenging part is probably the platform version differences.
I'm going to squash and merge this into my branch now.
For future reference, please don't merge branches but rebase instead.

dmfs:
Optionally, if you want to see yourself as the author of these commits, you can squash your branch now and I'll just fast-forward the notification branch. Let me know.

christianlupus:
This is up to you. I am sorry, I messed with your history. I tried to get things clean again. I hope this saves you time not to have to worry about things.
If I made things worse, I put another branch in my forked repo called `dev/notification-merge`, that should contain the merged version from before

dmfs:
No worries, I would just have squashed the commits which would get rid of the merge commit. Anyway, this way it's fine too.

PR #751, opened 2019-01-25
 fix recursive updateNotification calls 

Unpublished:
#749 always updates notifications after transaction.
As updating notificaton queries the database it gets stuck in a loop.

Partly reverted #749 to fix this.

### DIFF
--- a/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
+++ b/opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
@@ -152,6 +152,11 @@ public final class TaskProvider extends SQLiteContentProvider implements OnAccou
     Handler mAsyncHandler;
 
     /**
+     * Boolean to track if there are changes within a transaction.
+     */
+    private boolean mChanged = false;
+
+    /**
      * This is a per transaction/thread flag which indicates whether new lists with an unknown account have been added.
      * If this holds true at the end of a transaction a window should be shown to ask the user for access to that account.
      */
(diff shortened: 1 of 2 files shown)


dmfs:
Ouch! Thanks for finding this so quickly. I don't know how I could miss that.

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
```diff
@@ -1298,7 +1313,10 @@ protected void onEndTransaction(boolean callerIsSyncAdapter)
     {
         super.onEndTransaction(callerIsSyncAdapter);
         Intent providerChangedIntent = new Intent(Intent.ACTION_PROVIDER_CHANGED, TaskContract.getContentUri(mAuthority));
-        updateNotifications();
+        if (mChanged)
```
dmfs: Doesn't `mChanged` have to be reset to `false` somewhere?

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
```diff
@@ -1298,7 +1313,10 @@ protected void onEndTransaction(boolean callerIsSyncAdapter)
...
+        {
+            updateNotifications();
```
Unpublished:
```suggestion
            updateNotifications();
            mChanged = false;
```

opentasks-provider/src/main/java/org/dmfs/provider/tasks/TaskProvider.java
Unpublished: Seems reasonable to me.

PR #798, opened 2019-05-25
Bump deps

Unpublished:
Part of #797 

Running this since a few weeks without problems.


Unpublished:
Okay I fixed the lint error caused by mixed support libs by migrating to AndroidX (which should be done anyway).

Will run this over the next few days, probably upgrade them further to their latest release (1.0.2) and report back.

dmfs:
Thanks a lot. I'll check it out and give it a try.

Unpublished:
Great! PR should now be in its final state.

`androidx.annotation` 1.0.2 is not relevant for opentasks and test_runner strictly depends on 1.0.0, so it cannot be upgraded yet.

Unpublished:
Thanks for merging!

Was about to report the same, I didn't run into issues with this branch as well.

Issue #741, opened 2019-01-18
Reconsider Inserting into Instances Table


dmfs:
While implementing recurrence support we've also added the ability to insert tasks directly into the instances table. This approach fails if the task is beyond the expansion window (which is currently one single instance) because even though an exception is created, no entry is created in the instances table.

The feature is be disabled for now but we need to check if it makes sense to re-enable it.
One way could be to create "protected" instances which are always kept during expansion, so the newly inserted instance will always we shown. However that doesn't make much sense until we expand more than just one task in advance.

Issue #462, opened 2017-10-18
Recurrence


dmfs:
This epic groups all issues related to recurring tasks.

steve-s-71:
@dmfs please can you share if this is getting addressed anytime soon? 

Many people seem to wait for this since quite some time. I have seen older comments saying "we are working on it" but that has obviously not materialized. It would be really good if you could share your plans and schedules in more detail, as it helps the affected users to decide between being patient a bit longer or moving away from this app.

MisterY:
I'll just mention this, as it may solve people's issues temporarily - the combination of the sync adapter (DAVdroid or CalDAV sync) + OpenTasks + aCalendar(+) works very well with recurring tasks. The hack for storing the recurrence info in Google Tasks, for example, is not used in this scenario so I assume that the native support for NextCloud recurrence is used. I haven't looked into details but am pretty happy with this combination as it gets the job done.
This would imply that the tasks storage, provided by OpenTasks, is functioning well and it is only the UI part that is missing?
Hope that having the workaround solution takes some pressure off.

steve-s-71:
I got this combination setup earlier (precisely it was the paid app „aCalendar+ Kalender & Tasks“ [click](https://play.google.com/store/apps/details?id=org.withouthat.acalendarplus&hl=de)). 

But the app messed up most of my recurring tasks in a very odd way: After a Sync was triggered manually or by time (no modification of any task, just Sync) existing recurring tasks got modified. For example due date was altered, or recurrence was changed from weekly to daily, etc).

MisterY:
Thanks for the tip. Will pay close attention.
I've been using it for about a week or so and have not noticed any issues so far. I have a few mobile devices synchronizing the same calendars. The infrastructure is NextCloud 12 with Calendar 1.5.6, I think.
There could be multiple causes for the issue you describe - client app, sync adapter (I've first used CalDAV Sync, now I'm on DAVdroid as it provides CardDAV and CalDAV in one package), server infrastructure.

steve-s-71:
Maybe my problems is an interoperability issue:

Currently I have multiple Android devices running OpenTasks, synchronizing to the Nextcloud Server via [DavDroid](http://davdroid.com) and PC desktops running Outlook, synchronizing to the Nextcloud Server via [caldavsynchronizer](http://caldavsynchronizer.org/).

To this point my setup is working with recurring as well as non recurring tasks (of course I must not edit  recurring task in OpenTasks as that app does not support recurring tasks).

[...]

PR #814, opened 2019-07-26
Add initial recurrence support, implements #462

dmfs:
This commit adds initial recurrence support in that recurring instances are handled correctly when instances are edited or deleted.
In order to support this, the UI now operates on the Instances table rather then the tasks table. The instances table now supports deletes, and updates like the tasks table with the difference that all these operations only affect the specific instance that's addressed. All the operations on the instances table are converted into operations on a task by creating exception instances and rdates or exdates.

Inserting instances is currently not possible because only a fixed number of instances are expanded. This means an instance inserted beyond the expansion window would still not show up in the instances table.
See https://github.com/dmfs/opentasks/issues/741

There is still room for improvement in that we could split of completed tasks and modify recurrence rules when delting instances from the start or end of the series. That's left to future updates.
Also there is no UI yet to create or edit recurrence properties itself.

At present there is only one upcoming instance expanded, so you can only see the "current" instance (and all completed ones). This may/will change in future but may require some UI changes as well.

In order to support this commit a number of tests have been implemented which test creating and altering recurring instances.

Currently an exception is created when a recurring instance is modified which wasn't an exception before. This approach is RFC 5545 compliant but not supported by Apple clients. This will be fixed in future updates.

### DIFF
--- a/opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
+++ b/opentasks-contract/src/main/java/org/dmfs/tasks/contract/TaskContract.java
@@ -776,8 +776,8 @@ public final class TaskContract
         String ORIGINAL_INSTANCE_ID = "original_instance_id";
 
         /**
-         * The time in milliseconds since the Epoch of the original instance that is overridden by this instance or <code>null</code> if this task is not an
-         * exception.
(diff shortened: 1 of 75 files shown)


dmfs:
Ok, this has been pending way to long. I'll merge this one and we'll see how it goes.

PR #828, opened 2019-08-25
fix adaptive icon being too small

### LINKED ISSUE #827: Provide larger icon in app overview
Icon of this app in the app overview in settings is too small, see screenshot.

Go in Settings to App and notifications and show all apps.

tasks

Installed the app via F-Droid. See also https://github.com/xbmc/Kore/issues/670 for similar issue and possible fix.

javl:
Fixes the adaptive icon being too small by using 108dp instead of the old 24dp, as per this official Android guide: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive

Note: according to the guide 108dp is the default for Android >7.1. I do not have access to any devices with version <=7.1 so did not test on those.
 
Should close #827

### DIFF
--- a/opentasks/src/main/res/drawable/ic_opentasks_adaptive_fg.xml
+++ b/opentasks/src/main/res/drawable/ic_opentasks_adaptive_fg.xml
@@ -1,9 +1,9 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:aapt="http://schemas.android.com/aapt"
-        android:height="24dp"
+        android:height="108dp"
         android:viewportHeight="28.575"
         android:viewportWidth="28.574999"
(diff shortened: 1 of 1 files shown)


dmfs:
Oh, I was always wondering why that happened. Thanks for fixing this. I still need to test this on a pre 7.1 device before merging though.

javl:
Just realized I still have a Nexus 7 tablet with Android 5.1.1

First image is the icon after installing the app from the store (google photos app shows as size reference), the second image is after the update. Seems the older version of Android use the .png files that are in the repo, instead of the adaptive .xml file (the green blob is an edit I made to the png to be sure I was updating the image).

MVIMG_20190826_105623~2
MVIMG_20190826_110734~2

PR #833, opened 2019-08-28
Fix duplicate task after moving a task to a new list

dmfs:


dmfs:
There seems to be a problem with jitpack. I'll try again later.

Issue #834, opened 2019-08-28
Remove "afterKeys" and "ensureUpdates" from ContentSet


dmfs:
Neither of these things is really necessary anymore.

dmfs:
done

PR #835, opened 2019-08-28
Remove obsolte code, implements #834

dmfs:
It's no longer necessary to update all values which affect recurrince in the same update. The provider can handle updates of individual fields.


dmfs:
There seems to be a problem with jitpack. I'll try again later.

Issue #454, opened 2017-10-14
Update relations when modifying `PARENT_ID` of a task.


dmfs:

lemonboston:
I suppose it make sense to do the existing relation clearing as part of this, right?

Having 2 internal operations:

- create relation:
  - add PARENT_ID to subtask in TASKS
  - remove all PARENT Relation Property from PROPERTIES for the subtask
  - add PARENT Relation Property

- remove relation:
  - remove PARENT_ID from subtask in TASKS
  - remove all PARENT Relation Property from PROPERTIES for the subtask

Does this sound good?
I suppose we can use OpenTaskPal for these 'internal' operations as well, right?
The question is how to make sure they are not used in the app.

dmfs:
I'm leaning towards keeping the current behavior, having `PARENT_ID` read-only. So instead of this we should make sure you can't write to `PARENT_ID`

dmfs:
Done.

PR #876, opened 2019-11-11
Document that task column parent_id shouldn't be set directly (see #454)

rfc2822:


dmfs:
Updating `PARENT_ID` directly can be a nice feature when it comes to creating a child task in the UI. In such case we can just pass the `PARENT_ID` field to the editor and don't need to perform another provider request afterwards. The content provider would then take care of updating the relations (as stated in #454).
However, it doesn't make much sense for a sync adapter to update this field directly. So how about me make it read-only for sync adapters and keep it writable for the UI and other non-sync-adapter clients?

rfc2822:
Hi,

Thanks for the quick reply! That's the thing: At the moment, aCalendar+ sets `parent_id`, but doesn't create a relation. So there are currently two independent methods of setting parent relations:

1. with `parent_id` (used by aCalendar+)
2. with a `Relation`

As far as I understand #454, it should be enough for sync adapters to sync the `Relation`. But at the moment, this is not enough and both methods have to be implemented separately (because aCalendar+ sets `parent_id` and OpenTasks does not create a relation from it yet).

So I'd just like to know how OpenTasks is intended to be used at the current state of implementation: is it OK that apps like aCalendar+ set `parent_id` without creating a `Relation` (→ sync adapters must implement both methods to detect relations) or shall they create a `Relation` instead?

dmfs:
My idea is that sync adapters always update relations, because that's probably the easier way. In particular, the parent task may not be known (only its UID), in which case updating `parent_id` won't work. A relation doesn't require the parent to exist in the task list. Another advantage is that you don't have to care about syncing the tasks in the right order (parent before child).

The UI and 3rd party apps like aCalendar can use either, relations or `parent_id`, whatever is easier. The provider is supposed to update the relations when the `parent_id` is updated (that's what #454 is about). Of course we also need a migration which creates relations for every `parent_id` that's already in the database.

So yes, aCalendar is correct in setting the `parent_id` but OpenTasks still has to implement this feature. I think this shouldn't be hard to do. I'll include this in the next release.

rfc2822:
I see, thanks for the clarification and your efforts!

Issue #617, opened 2018-01-07
Split recurring tasks when modifying the first instance


dmfs:
It's probably a good idea to split a recurring tasks when overriding the first instance. After the split there would be a single task for the override and a recurring task starting at the next instance (if there is any). This way we don't have to deal with all the old instances of completed tasks. Also some clients seem to support this model only (i.e. Apple clients).
The drawback is, we loose the information about the recurrence, i.e. the current instance would appear as a single task, not a recurring one. One solution to this could be to add a special relation which identifies this task as a part of a recurrence set.
TODO: decide which of these tasks is the result of the old task being modified and which one is the newly created one.

dmfs:
For starters, let's only split off completed instances and see how this works out with Apple and other clients.

dmfs:
Done.

PR #899, opened 2019-12-06
Detach completed instances, #617

dmfs:
In order to support simple recurrence models and also keep long running tasks small we detach completed instances at the beginning of a series into separate task instances.


dmfs:
Looks like a couple of comments are outdated. They have to be fixed before merging this.

Issue #875, opened 2019-11-10
Support markdown lists with dash for checklists


unclev:
As _OpenTasks_ is often used with _Nextcloud_ there is an inconsistency in representing list (in Nextcloud description field), and _OpenTasks_' checklist.

For eg., the following list created in _Nextcloud_

```
- [x] Show the standard format usage (supported by Nextcloud): copy markdown
- [ ] How it is shown in the Android Tasks app: take image
```

renders pretty much the same as on GitHub

> - [x] Show the standard format usage (supported by Nextcloud): copy markdown
> - [ ] How it is shown in the Android Tasks app: take image

in _OpenTasks_ (v1.2.3 from _F-Droid_) goes into Description as plan text.

And vice versa, _OpenTasks_ checklist does not render to list in _Nextcloud_.

> [x] Show the standard format usage (supported by Nextcloud): copy markdown
>
> [ ] How it is shown in the Android Tasks app: take image

I have not seen a markdown flavour or renderer that treats `[ ]` and `[x]` as check list item, so it is probably worth treating `- [ ]` and `- [x]` as checklist items as well. (`[ ]` and `[x]` for backward compatibility).

For new list items on an existing list I suggest _OpenTasks_ to remember the last synchronization result, and to follow the convention used in the source.

This request is likely competes with #323, and a kind of checklist format extension similar to #265.

dmfs:
Makes sense to me. I don't see a need for retaining the conventions used for a specific source. This would only make it more complicated for little benefit (if any at all).

dmfs:
This should be fixed in the upcoming release.

PR #901, opened 2019-12-23
Render Checkboxes using more common Markdown style, implements #875

dmfs:
In addition this changes the check list UI. Instead of splitting the description into a description and check list part, we treat each paragraph either as text paragraph or a checkable item. Checkable items and text items can be converted into each other.


romainreignier:
I have tested it, it's a nice improvement. Thanks! :+1:

PR #909, opened 2020-01-26
One more PL update

Valdnet:
Update information translation.


dmfs:
Would you mind merging this with #908 on your end first? Thx.

Valdnet:
@dmfs Could it be like that? Thx.

Valdnet:
@dmfs Can you delete `#3` in the commit name? Thx.

PR #918, opened 2020-02-11
Update PL

Valdnet:
Translation correction.


dmfs:
Will merge this once I verified that f-droid has picked up the latest release.

Valdnet:
@dmfs I see that f-droid has chosen the latest version 😉.

Issue #17, opened 2013-08-09
Widget is not updated when task become "current tasks"


dmfs:
The widget should be updated if the start date or the due date of a task has passed.

The widget should determine the next upcoming start date or due date and set an alarm.

wweich:
I don't know if I should create a new issue or reuse this one, but I have this exact problem on my Nexus 5 and Nexus 7 (Android 5.1, stock).
I get notifications for start and due date but the widget only updates on data change (new/edit/delete) but not when the start date or due date of a task passes.
It's the same on the default Google Now Launcher and Action Launcher 3.
Killing the Launcher will (of cause) also update the widget.

dmfs:
@wweich we'll look into that again.

Trogel:
I can still reproduce this issue, im my case on LineageOS 16.0-20200309 (Android 9) using a Honor 5X.

In addition, a task in the widget does not change its color to red when it becomes due, unless an update is triggered e.g. due to synchronization or by manually scrolling down and up through the list of tasks. At the same time, notifications work as expected.

I looked into the code and noticed that the `TaskListWidgetUpdaterService` only listens for `ACTION_TIMEZONE_CHANGED` and `ACTION_TIME_CHANGED`, whereas the notifications rely on `ACTION_BROADCAST_TASK_DUE` and `ACTION_BROADCAST_TASK_STARTING`. (According to [an answer on StackOverflow](https://stackoverflow.com/questions/7912885/action-time-changed-or-action-date-changed-cant-make-them-work/19112483#19112483) `ACTION_TIMEZONE_CHANGED` and `ACTION_TIME_CHANGED` "are triggered only when the user manually changes time or date, and they are not automatically triggered by the system".)

Therefore I changed `ActionService` to also update all widgets when receiving `ACTION_BROADCAST_TASK_DUE` or `ACTION_BROADCAST_TASK_STARTING`. For me this change resolves the issue. I'll provide a pull request soon.

dmfs:
should be fixed now.

PR #925, opened 2020-03-22
Update widgets when task starts or becomes due

Trogel:
This is an attempt to fix issue #17.

I hope it is not too bad style to talk to the widgets from the notification-related `ActionService`.

### DIFF
--- a/opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
+++ b/opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
@@ -25,6 +25,7 @@ import android.net.Uri;
 import org.dmfs.android.contentpal.RowDataSnapshot;
 import org.dmfs.tasks.contract.TaskContract;
 import org.dmfs.tasks.homescreen.TaskListWidgetProvider;
+import org.dmfs.tasks.homescreen.TaskListWidgetProviderLarge;
 import org.dmfs.tasks.R;
 
 
@@ -44,7 +45,13 @@ public final class UpdateWidgetsAction implements TaskAction
     public void execute(Context context, ContentProviderClient contentProviderClient, RowDataSnapshot<TaskContract.Instances> rowSnapshot, Uri taskUri)
     {
         AppWidgetManager appWidgetManager = AppWidgetManager.getInstance(context);
-        final ComponentName componentName = new ComponentName(context, TaskListWidgetProvider.class);
+        notifyTaskListDataChanged(appWidgetManager, new ComponentName(context, TaskListWidgetProvider.class));
+        notifyTaskListDataChanged(appWidgetManager, new ComponentName(context, TaskListWidgetProviderLarge.class));
+    }
+
+
+    private void notifyTaskListDataChanged(AppWidgetManager appWidgetManager, ComponentName componentName)
+    {
         final int[] appWidgetIds = appWidgetManager.getAppWidgetIds(componentName);
         appWidgetManager.notifyAppWidgetViewDataChanged(appWidgetIds, R.id.task_list_widget_lv);
     }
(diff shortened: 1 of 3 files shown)


opentasks/src/main/java/org/dmfs/tasks/homescreen/TaskListWidgetUpdaterService.java
```diff
@@ -233,7 +233,7 @@ public RemoteViews getViewAt(int position)
                 row.setTextViewText(android.R.id.text1, mDueDateFormatter.format(dueDate, DateFormatContext.WIDGET_VIEW));
 
                 // highlight overdue dates & times
-                if ((!dueDate.allDay && dueDate.before(mNow) || dueDate.allDay
+                if ((!dueDate.allDay && Time.compare(dueDate, mNow) <= 0 || dueDate.allDay
```
dmfs: This looks unrelated to the issue. Why did you make this change?

opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+/**
+ * A {@link TaskAction} that updates the widgets.
+ *
+ * @author Trogel
+ */
+public final class UpdateWidgetsAction implements TaskAction
+{
+    private final static String TAG = "UpdateWidgetsAction";
```
dmfs: Please remove unused field.

opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
```diff
@@ -0,0 +1,53 @@
...
+    public UpdateWidgetsAction()
+    {
+    }
+
+
+    @Override
+    public void execute(Context context, ContentProviderClient contentProviderClient, RowDataSnapshot<TaskContract.Instances> rowSnapshot, Uri taskUri)
+    {
+        AppWidgetManager appWidgetManager = AppWidgetManager.getInstance(context);
+        final ComponentName componentName = new ComponentName(context, TaskListWidgetProvider.class);
```
dmfs: Currently we have two widget providers (mostly for backwards compatibility). Please ensure to update both.

dmfs:
Thanks for the PR. Nice solution. I don't have any issues with using the notification actions. This mechanism is sort of proof of concept for a more generic plug-in mechanism. So it's nice to see we can use it for other use cases.

opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
Trogel: Oh, that was a relict of some debug logging. Removed.

opentasks/src/main/java/org/dmfs/tasks/actions/UpdateWidgetsAction.java
Trogel: Sorry, did not see the other. Hope I got it right.

opentasks/src/main/java/org/dmfs/tasks/homescreen/TaskListWidgetUpdaterService.java
Trogel: While working on this fix I observed some cases where the update was processed right within the second of the due date. I noticed that in those cases the task did not turn red but remained white in the widget. That seemed wrong. The reason was that dueDate was not before but exactly equal to mNow – at least in `android.text.format.Time`'s second precision.

Therefore I changed this comparison essentially from `dueDate < mNow` to `dueDate <= mNow`. Sorry, if the commit message was not clear about my intention.

Issue #443, opened 2017-10-09
Subtasks


dmfs:

dugite-code:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=53184789)](https://www.bountysource.com/issues/53184789-subtasks)

vinzv:
Added another 20$.

dschmidt:
And 35$ more.

varac:
Is this anywhere on the roadmap ? It would be awesome to have !

Nik-Sch:
I merged subtasks into the master branch and added some functionalities to the subtasks list in the detailedView. Most of the changes feel rather hacky but definitely work well. I will probably continue to improve the subtask integration but that's what I've got for now :)
3bcf760286e3fd1a6c40b7cd8928458dfd0a4e38

dschmidt:
So you just merged the old PR to master or what exactly did you do? When I click that link, I don't see any branch or PR information and the commit doesn't seem to be in master either.

When you have anything to test, let me know.

Nik-Sch:
It's a link to a commit in my fork where I did the changes. ([1.2.4-subtasks](https://github.com/Nik-Sch/opentasks/releases/tag/1.2.4-subtasks))
Besides the merging, I did a lot of changes to the behavior of the subtasks in the detailedView. However, I didn't create a PR because
1. I plan on doing further changes.
2. I have no experience in OpenTasks infrastructure and am pretty sure I broke coding guidelines.

PR #931, opened 2020-04-07
Subtask aded in detailedView

Nik-Sch:
As mentioned in #443, I made some changes to the PR #469 including quicker adding/checking/deleting of subtasks


dmfs:
First of all, thanks for your efforts. I appreciate that. I really like the idea of using quick add to add subtasks.
Here are a few thinks I'd like to get sorted before merging your changes
* I've rebased the original subtasks PR on the current master, you might want to rebase again
* I still want to apply a few more changes to the original PR, in particular support for subtasks should be specified by the task model, so we can support models which don't support subtasks
* I want to add some unit tests for the opentaskspal stuff
* I'm going to merge the original PR in smaller chunks

I hope to get this sorted over the holidays. In the meantime I have only a few comments on your PR.

opentasks/src/main/java/org/dmfs/tasks/QuickAddDialogFragment.java
```diff
@@ -77,9 +79,10 @@
     /**
      * The maximum time to add for the first time the "Task completed" info is shown.
      */
-    private final static int COMPLETION_DELAY_MAX = 1500; // ms
+    private final static int COMPLETION_DELAY_MAX = 1000; // ms
```
dmfs, line 82: I'd like to keep this delay a bit longer so new users will notice the confirmation. Please note that the actual delay gets shorter with every confirmation shown.

opentaskspal/src/main/java/org/dmfs/opentaskspal/rowsets/Subtasks.java
```diff
@@ -36,11 +38,12 @@
...
 {
 
+    @SuppressWarnings("unchecked")
     public Subtasks(@NonNull View<Tasks> view,
-                    @NonNull Projection<? super Tasks> projection,
+                    @NonNull Projection<Tasks> projection,
                     @NonNull RowReference<Tasks> parentTask)
     {
-        super(new QueryRowSet<>(view, projection, new ReferringTo<>(Tasks.PARENT_ID, parentTask)));
+        super(new QueryRowSetSorting<>(view, projection, new ReferringTo<>(Tasks.PARENT_ID, parentTask), new Present(Tasks.STATUS + ", " + Tasks.TITLE)));
```
dmfs, line 46: The custom `QueryRowSetSorting` is not necessary. You can apply the `Sorted` decorator to `view` giving you the same result. Check this out:
```
 super(
    new QueryRowSet<>(
        new Sorted<>(Tasks.STATUS + ", " + Tasks.TITLE, view),
        projection,
        new ReferringTo<>(Tasks.PARENT_ID, parentTask)));
```

opentasks/src/main/java/org/dmfs/tasks/QuickAddDialogFragment.java
Nik-Sch: Oh, I did not notice that, that's fine

opentaskspal/src/main/java/org/dmfs/opentaskspal/rowsets/Subtasks.java
Nik-Sch: Uh, that's awesome, thank you, I will adjust these changes tomorrow

Nik-Sch:
I rebased from the original PR #469,  included your suggestions and have a few further ideas/bugs:
1. Sometimes the subtasks view is loaded twice (see image). I don't really grasp why this could happen, maybe you have an idea.
2. Maybe the finished subtasks or even the percentage of subtasks should influence the parent's completed percentage, shouldn't it? Currently this happens with the markdown checkboxes, I don't know a good way to integrate both. Maybe one could average both, the "description percentages" and the "subtasks percentages" and if both are >0 average them for the parent's percentage and if only one is >0 take that one.
3. Another idea I had was to show the amount of subtasks a task has in the ListView but I was unsure about how to integrate it without overloading the list

two subtasks

dmfs:
Please see my comments regarding your points:
1. That's one of the reasons why the original PR needs some more work
2. I'd leave that to a separate story. This can easily get complicated. For instance, in project management or software development it's common that (sub-)tasks have different "weights" (e.g. story points).
3. Let's leave that to a separate story as well. This might need some support from the provider, e.g. we count the number of subtasks upon insert, so we don't need to do that every time when we show a task. Once we drop `ExpandableListView` and switch to a `RecyclerView` we could consider allowing to expand subtasks in the list.

I'd also like to reconsider deleting subtasks when deleting the parent. This may not always be the best choice. It might be better to ask the user whether subtasks should be kept and converted to "top-level" tasks or if they should be deleted too.

Nik-Sch:
Ok, all of your points sound reasonable. Is there something I could further do?

Issue #963, opened 2020-12-20
Not identifying and rejecting tasks titled all whitespaces as empty task


Alan052918:
When creating a new task in task edit view, leaving the title blank on save would trigger a warning prompt on the bottom of the screen that says "Empty task not saved.". However, filling the title with whitespaces wouldn't trigger the prompt and the blank-titled task is not killed. I propose that whitespace-titled tasks should be ruled as illegal and treated as actual "empty" tasks.

<img width="429" alt="截屏2020-12-21 上午1 01 41" src="https://user-images.githubusercontent.com/39738304/102719935-e637bf80-432b-11eb-825a-bfad6c0a3aff.png">

PR #964, opened 2020-12-20
Fix issue #963

Alan052918:


opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
```diff
@@ -678,7 +678,7 @@ public void saveAndExit()
 
         if (mValues.isInsert() || mValues.isUpdate())
         {
-            if (TextUtils.isEmpty(TaskFieldAdapters.TITLE.get(mValues)))
+            if (TextUtils.isEmpty(TaskFieldAdapters.TITLE.get(mValues)) || TaskFieldAdapters.TITLE.get(mValues).trim().equals(""))
```
h3xx, line 681: I think changing to `.trim().isEmpty()` would be more performant; it simply checks if the `length` is 0.

String.isEmpty was introduced in Java 6.

opentasks/src/main/java/org/dmfs/tasks/EditTaskFragment.java
```diff
@@ -714,7 +714,7 @@ public void saveAndExit()
                 }
             }
 
-            if (!TextUtils.isEmpty(TaskFieldAdapters.TITLE.get(mValues)) || mValues.isUpdate())
+            if ((!TextUtils.isEmpty(TaskFieldAdapters.TITLE.get(mValues)) && !TaskFieldAdapters.TITLE.get(mValues).trim().equals("")) || mValues.isUpdate())
```
h3xx, line 717: I think changing to `.trim().isEmpty()` would be more performant; it simply checks if the `length` is 0.

h3xx (changes requested):

